PR #1, opened 2026-07-05
Generate the dark logo variants and show the logo more widely

Xexanos:
## Summary

- **Logo colors in the app**: verified — the Compose theme already reuses the logo palette exactly (`#A93B28` copper primary, `#4F6B35` ash-green secondary in `ui/theme/Color.kt`), so no theme changes were needed.
- **Generator** (`tools/logo`): now derives the dark-background artwork instead of the manually assembled proposal. The frame swaps to `#8CAB64` and the mark's copper ramp shifts two steps toward its light end (body `#CF7B60`, tail fade `#D98C71 → #F8DCC6`), so the tail keeps **fading** on dark backgrounds rather than alternating between two tones. New outputs: `ratatoskr-logo-dark.svg`, `ratatoskr-lockup-dark.svg`, and the app's vector drawables (`drawable/ratatoskr_logo.xml` + `drawable-night/ratatoskr_logo.xml`) — the in-app logo now shares a single source with the SVGs, and the app automatically shows the dark variant in dark theme.
- **Sign-in screen**: shows the logo (120 dp) above the form, matching the Connect screen's pattern.
- **README**: opens with the logo (dark-mode aware via `<picture>`) and explains the color story (copper squirrel, ash-leaf green for Yggdrasil) in "About the name".
- Logo docs (`docs/logo/README.md`) updated accordingly; also fixed the stale `generate_logo.py` reference.

## Verification

- `gradlew :tools:logo:run` regenerates everything; the light SVGs stayed **byte-identical** (refactor is regression-free) and a second run produces no diff (deterministic).
- `gradlew :app:assembleDebug :app:testDebugUnitTest` — build green, unit tests pass.
- `drawable-night/ratatoskr_logo.xml` contains only the dark palette (no `#4F6B35` / `#A93B28` / `#C56A4F`).


Xexanos:
Second commit adds the accessibility check suite and corrects the logo story (tail = sound waves fading with distance, not motion).

**Accessibility findings** (from a one-off strict run at WARNING severity — the committed suite runs at the default ERROR severity and is green):

| Screen | Finding | Check |
|---|---|---|
| Sign-in (both states) | The "Sign in" button's speakable text is identical to the "Sign in" heading — indistinguishable for TalkBack users | `DuplicateSpeakableTextCheck` |
| Now playing (playing/paused) | Transport controls render emoji glyphs (`▶`/`⏸`/`⏹`) as text; their fixed emoji colors yield contrast ratios of 1.26 and 4.35 (min 4.50) and can't follow the theme — should become icons with `contentDescription`s | `TextContrastCheck` |

The suite is instrumented on purpose: the ATF checks walk the real accessibility node tree and pass vacuously on Robolectric (robolectric/robolectric#5642). A canary test with a knowingly broken UI fails loudly if the checks ever go blind.

Xexanos (commented):
Reviewed with a multi-agent pass (8 finder angles, each candidate independently verified against the code). 8 findings below as inline comments, most severe first: a launch-time crash loop after backup restore (KeystoreCrypto), swallowed coroutine cancellation (RatatoskrClient), a token-refresh race (ConnectionManager), plaintext password in saved instance state (SignInScreen), a sign-in screen that can't scroll under the keyboard, a seek slider that snaps back mid-drag, a SPEC/module-structure mismatch, and a silent-drift trap in the dark-palette recolor. Verified-but-dropped smaller items (over the 8-comment budget): the stale `generate_logo.py` reference in `docs/logo/ratatoskr-mark.svg:5`, the mark being regex-scraped in two places in GenerateLogo.kt (`vectorDrawable()` and `markOf`), DARK_COLORS being derivable from one ordered ramp list, and the splash hex literals in `values/themes.xml` duplicating Color.kt.

tools/logo/src/main/kotlin/io/github/xexanos/ratatoskr/tools/logo/GenerateLogo.kt
```diff
@@ -50,7 +59,26 @@ import kotlin.math.sin
...
 
-const val DARK = "#4F6B35"   // knot strand (E1 "Eschenlaub"; fallback E2 #3A5230)
+const val FRAME_LIGHT = "#4F6B35"   // knot strand (E1 "Eschenlaub"; fallback E2 #3A5230)
+const val FRAME_DARK = "#8CAB64"    // knot strand on dark backgrounds
+
+// Dark-background palette: the frame swaps to FRAME_DARK, and the mark's copper ramp
+// (body, then the tail fade inner -> outer) shifts two steps toward the light end, so
+// the tail keeps fading on dark backgrounds. Several colors are both source and target,
+// so recoloring must happen in a single pass over the fragment.
+val DARK_COLORS = mapOf(
```
Xexanos: `DARK_COLORS` hand-codes a "shift the copper ramp two steps toward the light end" rule as 7 interlocking hex pairs where several colors are simultaneously a source and a target. That aliasing is exactly why the comment has to warn about single-pass recoloring. The whole map is derivable from one ordered ramp:
```kotlin
val COPPER_RAMP = listOf("#A93B28","#C56A4F","#CF7B60","#D98C71","#E29D82","#EBAE93","#F1BEA4","#F5CDB5","#F8DCC6")
val DARK_COLORS = buildMap {
    put(FRAME_LIGHT, FRAME_DARK)
    COPPER_RAMP.dropLast(2).forEachIndexed { i, c -> put(c, COPPER_RAMP[i + 2]) }
}
```
I verified this reproduces your map exactly. It makes the "+2 steps" rule explicit and removes the risk of a mistyped pair silently breaking the tail's monotonic fade. Not blocking.

tools/logo/src/main/kotlin/io/github/xexanos/ratatoskr/tools/logo/GenerateLogo.kt
```diff
@@ -50,7 +59,26 @@ import kotlin.math.sin
...
+    "#C56A4F" to "#D98C71",   // tail fade, inner -> outer
+    "#CF7B60" to "#E29D82",
+    "#D98C71" to "#EBAE93",
+    "#E29D82" to "#F1BEA4",
+    "#EBAE93" to "#F5CDB5",
+    "#F1BEA4" to "#F8DCC6",
+)
+
+fun recolor(fragment: String): String =
+    Regex("#[0-9A-Fa-f]{6}").replace(fragment) { DARK_COLORS[it.value.uppercase(Locale.ROOT)] ?: it.value }
```
Xexanos: `recolor` silently passes through any color not in `DARK_COLORS` (`?: it.value`). Today every fill in the mark is covered, so the output is correct — but `ratatoskr-mark.svg` is explicitly hand-edited in Inkscape (its header calls it the source of truth), and the moment someone nudges or adds a tail color there, the dark SVGs and `drawable-night` will ship that path in its *light* color with no error and no failing check — invisible until someone eyeballs the dark logo. Consider failing loudly on an unmapped fill (require every `#rrggbb` found in the fragment to be a known key, modulo the background). Separately, `#[0-9A-Fa-f]{6}` has no trailing boundary, so a future 8-digit `#RRGGBBAA` fill would have its first 6 digits rewritten and the alpha left dangling; a `(?![0-9A-Fa-f])` guard would future-proof it.

tools/logo/src/main/kotlin/io/github/xexanos/ratatoskr/tools/logo/GenerateLogo.kt
```diff
@@ -295,6 +322,49 @@ fun wordmarkPaths(font: File): String {
...
+""".trimIndent() + "\n"
+
+fun vectorDrawable(knotD: String, mark: String, sc: Double, tx: Double, ty: Double): String {
+    fun fmt(v: Double) = "%.1f".format(Locale.ROOT, v)
+    fun path(indent: String, fill: String, evenOdd: Boolean, d: String) =
+        "$indent<path android:fillColor=\"$fill\"" +
+            (if (evenOdd) " android:fillType=\"evenOdd\"" else "") +
+            " android:pathData=\"$d\"/>\n"
+
+    val markPaths = Regex("<path[^>]*/>").findAll(mark).joinToString("") { m ->
```
Xexanos: This `Regex("<path[^>]*/>")` scrape of the mark (plus the per-attribute `fill=`/`d=` regexes just below) is a second copy of the same ad-hoc SVG parsing already done by `markOf` at line 403. Both assume self-closing `<path .../>` tags with presentation attributes — but Inkscape (the documented editor for `ratatoskr-mark.svg`) routinely rewrites paths as `<path ...></path>` or moves fills into `style="fill:#..."`. If that happens, `markOf` silently yields fewer paths (no error path) while `vectorDrawable` would hit `error("mark path without fill")` — the SVG and the Android drawable would then diverge from the same source. Parsing the mark once into a shared `List<(fill, evenOdd, d)>` consumed by both the SVG composition and the drawable emitter would remove the duplication and the drift. Not blocking.

Xexanos (commented):
**Re-review against the corrected base (`scaffolding`).** My earlier review ran while this PR was still based on `main`, so its diff wrongly included the entire scaffolding tree; I've deleted those 8 inline comments. The real diff here is 17 files / ~313 lines: the dark-logo generation in `GenerateLogo.kt` (`DARK_COLORS`/`recolor`/`vectorDrawable`), the generated light+dark drawables, the new `AccessibilityChecksTest` + its dependency, `private`→`internal` on the preview composables so the test can reach them, and docs.

I read the full diff and found **no correctness bug in this PR's actual changes**. A few things I checked and cleared: the README `<picture>` fallback `docs/logo/ratatoskr-logo.svg` does exist at this head; the regenerated light drawable now carries scientific-notation floats (`2.01e-5`) in `android:pathData`, which is fine on `minSdk 26` (the platform `PathParser` handles exponents); `recolor`'s single-pass regex correctly handles the source-is-also-target aliasing in `DARK_COLORS`; and the `private`→`internal` widening is the standard way to expose previews to the androidTest friend module. 3 quality findings on the generator below.

**Note on the earlier findings:** the serious ones I posted before (backup-restore decrypt crash in `KeystoreCrypto`, swallowed `CancellationException` in `RatatoskrClient`, the refresh-token race in `ConnectionManager`, plaintext password / missing password keyboard, sign-in not scrollable, now-playing slider resetting mid-drag) are all in `scaffolding` — the base — not introduced here. They looked real; if there's an open PR for the scaffolding branch they belong there. Happy to re-file them against it.

tools/logo/src/main/kotlin/io/github/xexanos/ratatoskr/tools/logo/GenerateLogo.kt
Xexanos: Another Option would be to always generate both images and keep the parent image monochrome.

Xexanos: Went with your alternative here (generate both, monochrome parent). `ratatoskr-mark.svg` is now a monochrome source: each path carries a `data-ramp` index (its position in `COPPER_RAMP`) plus a `#808080` placeholder fill for the Inkscape preview. The tool paints the mark for both themes straight from the ramp — light = `ramp[i]`, dark = `ramp[i + DARK_SHIFT]` — and the frame colour is a per-theme parameter. So there's no light→dark colour map (and none of its source-is-also-target aliasing) left to keep in sync: `DARK_COLORS` and `recolor()` are gone. Generated SVGs and drawables are byte-for-byte identical. c030fc5

tools/logo/src/main/kotlin/io/github/xexanos/ratatoskr/tools/logo/GenerateLogo.kt
Xexanos: Resolved as part of the ramp rework rather than by hardening `recolor()`: `recolor()` no longer exists, so there's no silent passthrough to guard. The mark's colour now comes from an explicit `data-ramp` index, and a missing / non-integer / out-of-range index throws during generation (`parseMark` + `COPPER_RAMP[i + shift]`), so a mis-indexed or newly-added mark path fails loudly instead of shipping in the wrong colour. The 8-digit `#RRGGBBAA` concern is moot too — the painter replaces the whole `fill="…"` value, not a bare 6-hex run. (An interim commit did add the fail-loud guard + `(?![0-9A-Fa-f])` boundary to `recolor()`; it was rebased out once B removed `recolor()` entirely.) c030fc5

tools/logo/src/main/kotlin/io/github/xexanos/ratatoskr/tools/logo/GenerateLogo.kt
Xexanos: Done — the mark is parsed once by `parseMark()` into `List<MarkPath>`, consumed by both the SVG composition (`paintMarkSvg`) and the drawable emitter (`vectorDrawable`), so the two can't diverge. The regex now matches an opening `<path …>` tag whether or not it's self-closed, so an Inkscape `<path></path>` rewrite is handled, and it fails loudly on a path with no `d` or no `data-ramp` index. The fill-in-`style=` case is now moot: after the ramp rework the mark's colour comes from its `data-ramp` index, not from a fill in the source at all. a4898da (single parse) / c030fc5 (index-based colour)

PR #2, opened 2026-07-05
Fix the accessibility findings and make WARNING the checks' default severity

Xexanos:
Stacked on #1 (base: `logo/dark-theme-icons`). Brings together the two accessibility-fix sessions, plus the screen-polish work they were built on (`ceca723`, from `scaffolding`), and raises the bar of the accessibility suite.

## Summary

- **Sign-in**: heading is now "Sign in to Audiobookshelf", so the "Sign in" button no longer reads identically to it for TalkBack (`DuplicateSpeakableTextCheck`).
- **Now playing**: the transport controls are vector icons (PlayArrow from material-icons-core; Pause/Stop hand-drawn on the 24dp grid instead of pulling in the extended-icons artifact), tinted by the theme and speaking as Play/Pause/Stop.
- **`AccessibilityChecksTest` now runs at WARNING severity by default** (previously ERROR), which adds the contrast checks. Running that in both light and dark mode surfaced two more findings, fixed here:
  - **Connect**: the URL field was *disabled* during certificate inspection/confirmation — Material renders disabled text at 38% alpha (contrast 3.05 on the dark background). It is now `readOnly`: legible, but unchangeable under the certificate being confirmed.
  - **Library/Speakers empty states**: the decorative emojis (📚/🔍/🔇) carry fixed emoji colors and were measured as text (ratio 3.58 in light mode). They now sit outside the semantics tree (`clearAndSetSemantics`), which also stops TalkBack from announcing them as stray words.

## Verification

- Full `AccessibilityChecksTest` suite (15 screens + canary) run on an API 36 emulator in **both light and dark mode**: green at WARNING severity.
- `:app:testDebugUnitTest` green.


.github/workflows/accessibility.yml
```diff
@@ -0,0 +1,70 @@
...
+#
+# The suite is instrumented on purpose — the Accessibility Test Framework needs the
+# real accessibility node tree, which only exists on a device (see the test's kdoc) —
+# so this boots a headless emulator on the KVM-enabled GitHub runner. Contrast is
+# theme-dependent, so the suite runs twice: once in light mode, once in dark mode.
+
+name: Accessibility
+
+on:
+  pull_request:
```
Xexanos, line 15: **`on: pull_request` has no `paths:` filter, so this 45-minute emulator job boots on every PR** — including README-only, `docs/**`, or F-Droid metadata changes that can't affect the a11y suite. Given the repo actively touches docs and fastlane metadata, that's a lot of wasted runner time and an irrelevant required check blocking those PRs.

Suggestion: scope the trigger, e.g.
```yaml
on:
  pull_request:
    paths:
      - 'app/**'
      - 'gradle/**'
      - 'gradle/libs.versions.toml'
      - '.github/workflows/accessibility.yml'
  workflow_dispatch:
```

.github/workflows/accessibility.yml
```diff
@@ -0,0 +1,70 @@
...
+        uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: 21
+
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      - name: Run the accessibility checks in light and dark mode
+        uses: reactivecircus/android-emulator-runner@v2
```
Xexanos, line 80: **No AVD snapshot caching** — the API-36 x86_64 AVD is created and cold-booted from scratch on every run (~3–6 min of the budget each time). `reactivecircus/android-emulator-runner` supports caching the AVD across runs: add an `actions/cache` step for `~/.android/avd/*` + `~/.android/adb*` keyed on the emulator params and set `force-avd-creation: false`, so subsequent runs restore a warm snapshot. (See the action's README "AVD caching" example.)

app/src/main/java/io/github/xexanos/ratatoskr/ui/speakers/SpeakersScreen.kt
```diff
@@ -102,30 +113,72 @@ private fun SpeakersContent(
...
+                Column(
+                    horizontalAlignment = Alignment.CenterHorizontally,
+                    verticalArrangement = Arrangement.spacedBy(8.dp),
+                    modifier = Modifier.padding(horizontal = 32.dp),
+                ) {
+                    // Decorative, like the library empty state: out of the semantics tree.
+                    Text(
+                        "🔇",
+                        style = MaterialTheme.typography.displaySmall,
+                        modifier = Modifier.clearAndSetSemantics {},
```
Xexanos, line 163: **Empty-state block is duplicated.** This `Box(fillMaxSize, Center)` → `Column(CenterHorizontally, spacedBy(8.dp), padding(horizontal = 32.dp))` → decorative emoji with `clearAndSetSemantics {}` + title + body is a near-verbatim copy of `EmptyLibrary` in `LibraryScreen.kt:201`; the comment here even says "Decorative, like the library empty state". The a11y idiom (`clearAndSetSemantics` on the glyph) is exactly the pattern this PR is trying to establish — if it's copy-pasted per screen, the next empty state will re-implement it and can silently omit the fix, reintroducing the TalkBack-noise/contrast finding.

Suggestion: extract a shared `EmptyState(emoji: String, title: String, body: String)` composable and call it from both screens (and future ones).

app/src/main/java/io/github/xexanos/ratatoskr/ui/connect/ConnectScreen.kt
```diff
@@ -125,53 +137,174 @@ private fun ConnectContent(
...
+        Spacer(Modifier.height(32.dp))
         OutlinedTextField(
             value = url,
             onValueChange = { url = it },
             label = { Text("Server URL") },
             singleLine = true,
+            // readOnly, not disabled: while inspecting/confirming the certificate the URL
+            // must stay legible (disabled text is 38% alpha and fails contrast checks),
+            // it just must not change under the certificate being confirmed.
+            readOnly = !(state is ConnectUiState.Idle || state is ConnectUiState.Error),
```
Xexanos: Minor/UX: `readOnly` fixes the 38%-alpha contrast issue and keeps the URL legible during certificate confirmation — good. But a `readOnly` `OutlinedTextField` still renders the full editable affordance (focusable border, floating label, tappable cursor), so visually it reads as "editable but unresponsive" rather than "locked to the certificate being confirmed." Since this is the TOFU confirmation step where the URL↔fingerprint binding matters, consider signalling the locked state (e.g. a lock trailing icon, or rendering the URL as plain non-editable text within the card) so the intent is conveyed. Not blocking.

Xexanos (commented):
Reviewed the a11y/CI changes. **Coverage caveat:** the correctness-focused finder pass (line-scan, removed-behavior, cross-file, conventions) was cut short by usage limits mid-run, so unlike PR #1 this review leans on my own read of the diff rather than the full verified sweep. I read the substantive code changes directly — the hand-drawn Pause/Stop `ImageVector`s (correct path data, built once as top-level vals), the WARNING-severity switch + canary, the `readOnly` URL logic, the reworded sign-in heading, and the emoji `clearAndSetSemantics` — and found no correctness bug. A few things I explicitly cleared: pulling `material-icons-core` for `PlayArrow` is fine and consistent (the code documents the deliberate core-vs-extended split, and Pause/Stop are hand-drawn precisely to avoid the extended artifact); the decorative-emoji `clearAndSetSemantics` is the correct WCAG treatment for decorative content, not a check-dodge. 4 quality findings below.

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
```diff
@@ -371,11 +380,41 @@ private fun CircleControl(
...
+private val StopIcon: ImageVector = materialIcon(name = "Ratatoskr.Stop") {
+    materialPath {
+        moveTo(6f, 6f)
+        horizontalLineToRelative(12f)
+        verticalLineToRelative(12f)
+        horizontalLineTo(6f)
+        close()
+    }
+}
+
```
Xexanos, line 417: Please use the import instead of pulling in the shapes. With proper Tree Shaking the initial size of the artifact should be irrelevant.

app/src/main/java/io/github/xexanos/ratatoskr/ui/speakers/SpeakersScreen.kt
```diff
@@ -155,7 +156,12 @@ private fun SpeakersContent(
                     verticalArrangement = Arrangement.spacedBy(8.dp),
                     modifier = Modifier.padding(horizontal = 32.dp),
                 ) {
-                    Text("🔇", style = MaterialTheme.typography.displaySmall)
+                    // Decorative, like the library empty state: out of the semantics tree.
+                    Text(
+                        "🔇",
```
Xexanos, line 161: please use material-icons instead of emojis.

.github/workflows/accessibility.yml
Xexanos: Addressed — added a `paths-ignore` denylist so this run is skipped for docs/metadata-only PRs (`**/*.md`, `docs/**`, `fastlane/**`, `.gitignore`, `LICENSE`).

I deliberately used a denylist rather than the suggested `paths:` allowlist: an allowlist that misses a UI-affecting path would silently skip the accessibility check, whereas the denylist only skips changes fully confined to docs/metadata and still runs for everything else — including new or unlisted paths. Same rationale applied to the CI workflow, so the two stay consistent.

.github/workflows/accessibility.yml
Xexanos: Good optimization — not done yet. Caching `~/.android/avd/*` + `~/.android/adb*` with `actions/cache` (keyed on the emulator params) and `force-avd-creation: false` would restore a warm snapshot and save the ~3–6 min cold boot each run. Tracked as a non-blocking follow-up.

app/src/main/java/io/github/xexanos/ratatoskr/ui/speakers/SpeakersScreen.kt
Xexanos: Agree. The empty-state block is duplicated between Speakers and Library (it started in the UI-polish pass). Extracting a shared `EmptyState(icon, title, body)` composable is the right move — it keeps the `clearAndSetSemantics` a11y idiom in one place so a future empty state can't silently reintroduce the TalkBack/contrast finding. Good follow-up, and it pairs naturally with the emoji → Material-icon point below (switch both screens at once).

app/src/main/java/io/github/xexanos/ratatoskr/ui/connect/ConnectScreen.kt
Xexanos: Fair, and thanks for flagging it non-blocking. During the TOFU confirmation the URL↔fingerprint binding is the whole point, so signalling the locked state — a lock trailing icon, or rendering the URL as plain text inside the card — would read better than a disabled-looking editable field. Noting as optional polish.

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
Xexanos: Reasonable preference, but the tree-shaking premise doesn't hold yet: R8/shrinking is currently **off** for the release build (see #3, finding #14 — the reflective Moshi client needs validated keep rules, or a switch to Moshi codegen, before it can be enabled). With shrinking disabled, adding `material-icons-extended` would ship the whole extended set (thousands of vectors) unshrunk — which is exactly what the two hand-drawn shapes avoid.

So the clean sequence is: enable R8 (the #14 follow-up), then replace `PauseIcon`/`StopIcon` with `Icons.Filled.Pause` / `Icons.Filled.Stop` from `material-icons-extended` and delete the custom vectors. Until then I'd keep the shapes. (`PlayArrow` stays from material-icons-core either way.)

app/src/main/java/io/github/xexanos/ratatoskr/ui/speakers/SpeakersScreen.kt
Xexanos: Agree for visual consistency and theming — a Material icon (e.g. `Icons.Default.VolumeOff` for the no-speakers state, and something like `MenuBook` / `Search` for the Library empty states) beats an emoji glyph. They're decorative (`clearAndSetSemantics {}`), so it's purely cosmetic with no a11y impact. Best done together with the shared `EmptyState` extraction above so both screens switch in one place. Note the same "extended set needs R8" caveat as the transport icons if any of these live outside material-icons-core.

.github/workflows/accessibility.yml
Xexanos: Done in 440f082 — added an `actions/cache` step for `~/.android/avd/*` + `~/.android/adb*` keyed on the emulator params (`avd-api36-x86_64-default-pixel_6`), a warm-up step that saves a snapshot on cache miss, and set `force-avd-creation: false` + `-no-snapshot-save` on the run step so subsequent runs boot from the restored snapshot instead of cold-booting.

app/src/main/java/io/github/xexanos/ratatoskr/ui/connect/ConnectScreen.kt
Xexanos: Done in 7c0b6c0 — the URL field now shows a lock trailing icon while it's `readOnly` (the same `Icons.Default.Lock` the certificate card uses), so during TOFU confirmation it reads as "held to the certificate being confirmed" rather than editable-but-unresponsive. Still `readOnly`, not disabled, so the contrast fix is unaffected.

app/src/main/java/io/github/xexanos/ratatoskr/ui/speakers/SpeakersScreen.kt
Xexanos: Done in #4 (`ux/design-and-guardrails`): both empty states now render Material icons (`Icons.Default.SpeakerGroup` for the no-speakers state, etc.) via `material-icons-extended` instead of the emoji glyphs. Referencing that PR so the emoji → icon migration lands in one place rather than being split across both.

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
Xexanos: Referencing #4 (`ux/design-and-guardrails`): that PR pulls in `material-icons-extended` and migrates the transport controls to the imported icons, which will replace these hand-drawn Pause/Stop shapes. Keeping the shapes here and doing the swap there so the extended-icons dependency is introduced in one place.

app/src/main/java/io/github/xexanos/ratatoskr/ui/speakers/SpeakersScreen.kt
Xexanos: Agreed the block is duplicated. Since this pairs with the emoji → Material-icon swap ("switch both screens at once"), and that migration is happening in #4 (`ux/design-and-guardrails`) — which rewrites both empty states to Material icons + string resources — extracting a shared `EmptyState` on the emoji version here would be throwaway and would conflict with #4. So the shared extraction belongs on top of #4's icon version rather than in this PR.

Xexanos: Done in #4 (`ux/design-and-guardrails`, 1e49b21): extracted a shared `EmptyState(icon, title, body)` in the `ui` package and routed both the Library and Speakers empty states through it, so the decorative-icon a11y idiom lives in one place.

PR #3, opened 2026-07-06
Scaffold the app: Compose UI, generated API client, brand, and tests

Xexanos:
Base of the stack (`scaffolding → logo/dark-theme-icons → a11y/warning-severity`). Brings the project from the brief to a working, tested Compose app skeleton.

## Summary

- **App scaffold**: Android Studio project + Compose navigation, one screen/module per SPEC §13 (Connect, Sign-in, Library, Speakers, Now Playing, Settings), all previewable without a running server.
- **`:core-network`**: a thin wrapper over the openapi-generator client generated at build time from the contract submodule (SPEC §4), with TOFU certificate pinning (`PinnedTrustManager`), bearer-token attach, and single-flight token refresh.
- **Tests**: unit tests for the auth plumbing (bearer interceptor, refresh authenticator), the contract↔domain mappers, and the client wrapper (SPEC §9).
- **Brand**: the Ratatoskr logo, palette, launcher icon, and the `:tools:logo` generator; screens polished into cards/icons/empty-states with a clearer hierarchy.
- **Docs**: SPEC and the token-rotation proposal aligned with the merged contract 1.1.0.

## Notes

- The contract is a git submodule; `submodules: true` is needed to build (CI and F-Droid both support this).
- The two stacked PRs on top add the dark logo variants (#1) and the accessibility suite + CI (#2).


core-network/src/main/java/io/github/xexanos/ratatoskr/network/persist/KeystoreCrypto.kt
```diff
@@ -0,0 +1,69 @@
...
+        return Base64.encodeToString(combined, Base64.NO_WRAP)
+    }
+
+    fun decrypt(encoded: String): String {
+        val combined = Base64.decode(encoded, Base64.NO_WRAP)
+        val iv = combined.copyOfRange(0, IV_LENGTH)
+        val ciphertext = combined.copyOfRange(IV_LENGTH, combined.size)
+        val cipher = Cipher.getInstance(TRANSFORMATION)
+        cipher.init(Cipher.DECRYPT_MODE, getOrCreateKey(), GCMParameterSpec(TAG_LENGTH_BITS, iv))
+        return String(cipher.doFinal(ciphertext), Charsets.UTF_8)
```
Xexanos: **Uncaught decrypt failure → launch-time crash loop.** `decrypt()` calls `cipher.doFinal(ciphertext)` with the key from `getOrCreateKey()`, which *silently mints a new key* when the alias is missing (line 47 lookup misses → `generateKey()`). If the persisted ciphertext no longer matches the key, `doFinal` throws `AEADBadTagException`, and nothing catches it: `TokenStore.authSession()`/`refreshToken()`/`currentAccessTokenBlocking()` all call `crypto::decrypt` bare, and `MainActivity.decideStartDestination()` runs `authSession()` inside `runBlocking` in `onCreate` — so it crashes on every launch until the user clears app data.

The trigger is reachable: `android:allowBackup="true"`, and `data_extraction_rules.xml` has an empty `<cloud-backup>` (no include/exclude), so on Android 12+ **all** app data — including the token DataStore — is backed up by default, while Keystore keys never transfer to a new device. (Note the two backup configs disagree: `backup_rules.xml` restricts pre-12 backup to `sharedpref` only, so DataStore isn't backed up there — worth reconciling.) Independent of backup, AndroidKeyStore keys can also be lost across some OS updates.

Fix: catch `GeneralSecurityException` in the `TokenStore` read paths and treat it as signed-out (clear the stored tokens), so a key/ciphertext mismatch logs the user out instead of bricking launch.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClient.kt
```diff
@@ -0,0 +1,154 @@
...
+    ): ApiResult<T> = try {
+        val response = call()
+        if (response.isSuccessful) {
+            val body = response.body()
+            if (body != null) ApiResult.Success(body)
+            else ApiResult.Failure(RatatoskrError.Unexpected(IllegalStateException("Empty response body")))
+        } else {
+            ApiResult.Failure(mapHttpError(response.code(), response.errorBody()?.string(), sessionEndpoint))
+        }
+    } catch (t: Throwable) {
```
Xexanos, line 157: **`catch (t: Throwable)` swallows `CancellationException`, breaking structured concurrency.** `mapThrowable` (line 141) walks the cause chain for TLS/cert types then falls to `Network`/`Unexpected` — it never rethrows `CancellationException`. So when a `viewModelScope` is cancelled mid-request (user navigates away, ViewModel cleared), the coroutine's `CancellationException` is caught here and returned as `ApiResult.Failure(Unexpected(...))`; the cancelled coroutine then continues and writes an error UI state instead of stopping silently. Same in `executeUnit` (line 127).

Fix: rethrow before mapping —
```kotlin
} catch (t: Throwable) {
    if (t is kotlinx.coroutines.CancellationException) throw t
    ApiResult.Failure(mapThrowable(t))
}
```

app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
```diff
@@ -0,0 +1,55 @@
...
+
+    fun setSessionActive(active: Boolean) = sessionActive.set(active)
+
+    /** The client for the currently trusted server, or null if none is configured yet. */
+    suspend fun client(): RatatoskrClient? {
+        val config = connectionStore.currentServerConfig() ?: return null
+        val fingerprint = connectionStore.fingerprint() ?: return null
+        val key = "${config.baseUrl}|$fingerprint"
+        val current = cached
+        if (current != null && cachedKey == key) return current
```
Xexanos: **Unsynchronized client cache can spawn two OkHttp stacks and defeat single-flight refresh.** `cached`/`cachedKey` are plain `var`s and `client()` does a check-then-act (`val current = cached; if (current != null && cachedKey == key) return current`) in a `suspend fun` with no `Mutex`/`@Volatile`. Two coroutines racing the first call (e.g. a poll loop + a screen right after sign-in) each see `null` and each build a client via `RatatoskrClientFactory.create()`. The refresh guard is per-instance — `TokenRefreshAuthenticator` holds `private val lock = Any()` (AuthInterceptors.kt:55) and is constructed fresh per `create()` (RatatoskrClientFactory.kt:71) — so two stacks can concurrently `POST /auth/refresh` with the same token. Per SPEC section 5 the server rotates the refresh token on every use, so the losing refresh sends a now-dead token and can invalidate the stored pair → unexpected sign-out. (`sessionActive` only suppresses refresh *during* playback; it doesn't prevent the duplicate-stack race.)

Fix: guard `client()` with a `Mutex`, or memoize the build as a single-flight `Deferred`, so at most one client exists per connection key.

Xexanos (commented):
These findings surfaced during the PR #1/#2 reviews while those PRs were temporarily based on `main` (so their diffs pulled in this scaffold). They belong here, on the scaffold PR. I **re-verified each against the current `scaffolding` head** — this is a targeted pass on the findings themselves, not a full review of all 80 files (happy to run that separately). Two earlier findings are already handled in this branch and I'm not re-filing them: the sign-in `Column` now has `verticalScroll`, and the password field now sets `KeyboardType.Password`. 4 confirmed + 1 minor below.

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
```diff
@@ -0,0 +1,454 @@
...
+    }
+    Spacer(Modifier.height(12.dp))
+    Box(Modifier.fillMaxWidth(), contentAlignment = Alignment.Center) {
+        StateChip(session.state)
+    }
+
+    // Push transport controls into the thumb zone (bottom third).
+    Spacer(Modifier.weight(1f))
+
+    var sliderPosition by remember(session.positionSeconds) {
```
Xexanos: **Seek slider snaps back under the user's finger.** `remember(session.positionSeconds)` keys the slider state on the polled position, and the ViewModel polls every `POLL_INTERVAL_MS = 5_000L` (line 140), replacing the session unconditionally. During playback each poll changes `positionSeconds`, invalidating the `remember` key mid-drag: `sliderPosition` resets to the server value and the thumb jumps back. If the user then lifts their finger, `onValueChangeFinished = { onSeek(sliderPosition.toDouble()) }` (line 244) seeks to the reset value rather than the dragged target.

Fix: track an `isDragging` flag (set in `onValueChange`, cleared in `onValueChangeFinished`) and only re-sync `sliderPosition` from the session while not dragging.

app/src/main/java/io/github/xexanos/ratatoskr/ui/auth/SignInScreen.kt
```diff
@@ -0,0 +1,231 @@
...
+    SignInContent(state = state, onSignIn = viewModel::signIn)
+}
+
+@Composable
+private fun SignInContent(
+    state: SignInUiState,
+    onSignIn: (String, String) -> Unit,
+) {
+    var username by rememberSaveable { mutableStateOf("") }
+    var password by rememberSaveable { mutableStateOf("") }
```
Xexanos: Minor: the password is held in `rememberSaveable`, so it's written to the saved-instance-state Bundle (persisted to disk on process death). The keyboard type is now correctly `Password`, so only this half of the earlier note remains. Consider plain `remember` for the password field — losing it across process death is the right trade-off for a credential, and keeps the plaintext out of on-disk saved state. Not blocking.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/Mappers.kt
```diff
@@ -0,0 +1,92 @@
...
+
+internal fun GenPlaybackState.toDomain(): PlaybackState = when (this) {
+    GenPlaybackState.playing -> PlaybackState.PLAYING
+    GenPlaybackState.paused -> PlaybackState.PAUSED
+    GenPlaybackState.buffering -> PlaybackState.BUFFERING
+    GenPlaybackState.stopped -> PlaybackState.STOPPED
+    GenPlaybackState.finished -> PlaybackState.FINISHED
+}
+
+internal fun GenSession.toDomain(): Session =
```
Xexanos, line 83: **The refresh-token rotation obligations from SPEC §5 (contract 1.1.0) are not implemented.** `GenSession.toDomain()` never reads `rotatedTokens`, and the domain `Session` has no field for it, so the rotated pair that the server includes in every playback response (`getCurrentSession`/`start`/`pause`/`resume`/`seek`, and the `stopSession` 200 body) is silently dropped. Consequences, all against SPEC §5:
- `RatatoskrClient.stopSession()` (line 89) uses `executeUnit` and discards the body, so the final rotated pair is lost — SPEC says "treats any 2xx from stopSession as success and adopts rotatedTokens if the body carries it."
- `TokenRefreshAuthenticator` suppresses `/auth/refresh` while a session is active (`refreshSuppressed` → returns null on 401, AuthInterceptors.kt:58), but nothing implements the required fallback ("a 401 during an active session first triggers an immediate getCurrentSession to pick up a rotated pair") or the adoption itself.

Net effect: during a long/active session the access token eventually rotates server-side, the app never adopts the new pair and won't refresh on its own, so playback gets stranded and the user is forced to re-authenticate. This is the biggest gap in the scaffold's auth layer — the suppression half is wired but the adoption half is missing. Adding a `rotatedTokens` field to `Session` + adopting it in every playback result (and in the stop body) is the core fix.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/tls/PinnedTrustManager.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    private val platform: X509TrustManager = defaultTrustManager()
+
+    override fun checkClientTrusted(chain: Array<out X509Certificate>?, authType: String?) {
+        platform.checkClientTrusted(chain, authType)
+    }
+
+    override fun checkServerTrusted(chain: Array<out X509Certificate>, authType: String?) {
+        try {
+            platform.checkServerTrusted(chain, authType)
```
Xexanos, line 44: **Once a fingerprint is confirmed, the pin is not enforced if platform validation passes** — `checkServerTrusted` returns immediately after `platform.checkServerTrusted(...)` succeeds (line 38) and only compares the pin in the `catch` branch. SPEC §6 says to "persist the confirmed fingerprint and pin it for all later connections; warn loudly if it ever changes," and this class's own doc claims "a changed certificate is rejected loudly." Neither holds when the server presents a publicly-trusted cert: a *different* publicly-trusted cert for the host (server switched certs, or an attacker with a mis-issued/rogue-CA cert) is accepted silently, with no re-confirm prompt. For the pure self-signed setup (platform fails → pin checked) it works, but SPEC §6 explicitly assumes "no public CA in the chain," so the platform-first shortcut weakens the stated guarantee.

Consider: once a pin exists, always compare the leaf fingerprint to it (regardless of platform outcome), and only fall back to platform validation when no pin is set yet. **Also: this trust manager and `Fingerprints` — the entire TOFU/MITM defense — have no tests at all**, so a regression that inverts the comparison would ship green. Add coverage for platform-trusted-accept, pin-match-accept, pin-mismatch-reject, null-pin-reject, and empty-chain-reject.

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
```diff
@@ -0,0 +1,454 @@
...
+
+class NowPlayingViewModel(
+    private val connectionManager: ConnectionManager,
+) : ViewModel() {
+
+    private val _uiState = MutableStateFlow(NowPlayingUiState())
+    val uiState: StateFlow<NowPlayingUiState> = _uiState.asStateFlow()
+
+    init {
+        viewModelScope.launch {
```
Xexanos: **The polling loop's lifecycle has two problems.** (1) `while (isActive) { refresh(); delay(5s) }` lives in `viewModelScope`, which is not stopped when the app is backgrounded — so it keeps hitting the server every 5s (and keeps `setSessionActive(true)`) while the user is off-screen, until the ViewModel is cleared. Consider pausing polling when the screen isn't in the RESUMED state. (2) `stop()` sets `stopped = true` but does **not** stop this loop, and `applySession` (line 102) rebuilds the state as `NowPlayingUiState(loading = false, session = session)` — discarding both `stopped` and `error`. So a poll tick that lands right after `stop()` (server still briefly reporting a session) flips `stopped` back to false, and any transient error set by `refresh()`/`control()` is wiped by the next successful poll within 5s. Net: the user can tap Stop and remain stranded on Now Playing, and control-failure messages can vanish before they're read. Cancel/scope the poll on stop, and have `applySession` preserve `stopped`/`error` (use `copy` rather than a fresh state).

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
```diff
@@ -0,0 +1,365 @@
...
+) : ViewModel() {
+
+    private val _uiState = MutableStateFlow(LibraryUiState())
+    val uiState: StateFlow<LibraryUiState> = _uiState.asStateFlow()
+
+    init { load(null) }
+
+    fun search(query: String) = load(query.ifBlank { null })
+
+    private fun load(query: String?) {
```
Xexanos: **Search has no debounce and no cancellation of the in-flight request.** `search()` calls `load(query)` on every keystroke; each `load` launches a new `viewModelScope` coroutine and replaces the whole `uiState` on completion, without cancelling the previous one. Responses can therefore land out of order (last-*completed* wins, not last-*typed*): type "dune" quickly and the earlier "du" request can resolve after "dune", leaving results for "du" under a search box reading "dune" — plus a burst of one request per character. Debounce the query (e.g. collect a `MutableStateFlow` with `debounce` + `flatMapLatest`) or at least cancel the prior job before launching the next.

app/src/main/java/io/github/xexanos/ratatoskr/MainActivity.kt
```diff
@@ -0,0 +1,55 @@
...
+                }
+            }
+        }
+    }
+
+    /**
+     * Launch routing (SPEC section 13): no trusted server → connect; no stored tokens →
+     * sign-in; otherwise the library. Reads are one-shot and fast (DataStore caches).
+     */
+    private fun decideStartDestination(container: AppContainer): String = runBlocking {
```
Xexanos: **`decideStartDestination` runs `runBlocking` on the main thread inside `onCreate`.** It performs three sequential DataStore reads (`currentServerConfig()`, `fingerprint()`, `authSession()`) plus a synchronous Android Keystore AES/GCM decrypt — all on the UI thread before `setContent`. The comment says DataStore caches, but on a cold start nothing is cached yet: each `.first()` collects from disk and the Keystore call is blocking JNI. On a slow device this drops launch frames or trips an ANR. Consider a lightweight loading destination while the start route is resolved off the main thread (e.g. decide inside a `LaunchedEffect`/splash state), rather than blocking `onCreate`.

app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
```diff
@@ -0,0 +1,55 @@
...
+            tokenStore = tokenStore,
+            sessionActive = { sessionActive.get() },
+        ).also {
+            cached = it
+            cachedKey = key
+        }
+    }
+
+    /** Drop the cached client after the server or certificate changed. */
+    fun invalidate() {
```
Xexanos, line 67: **Replaced OkHttp clients are never shut down → thread/FD leak.** `invalidate()` (and the cache-miss replacement in `client()`) drops the old `RatatoskrClient` by nulling the reference, but `RatatoskrClientFactory.create()` builds OkHttp stacks that each own a dispatcher `ExecutorService` and a `ConnectionPool` with keep-alive threads. `CertificateInspector.inspect()` correctly tears its client down (`dispatcher.executorService.shutdown()` + `connectionPool.evictAll()` in a `finally`), but `ConnectionManager` does not. Every re-trust / server change / fingerprint change during a process lifetime abandons a live thread pool and pooled sockets until GC finalizes them. Shut down the previous client before replacing it.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/Mappers.kt
```diff
@@ -0,0 +1,92 @@
...
+            progress = progress?.toDomain(),
+        ),
+        description = description,
+        narrator = narrator,
+    )
+
+internal fun GenLibraryItemPage.toDomain(): LibraryPage =
+    LibraryPage(items = items.map { it.toDomain() }, nextCursor = nextCursor)
+
+internal fun GenPlaybackState.toDomain(): PlaybackState = when (this) {
```
Xexanos, line 75: **`GenPlaybackState.toDomain()` can never produce `PlaybackState.UNKNOWN`.** The domain enum (Models.kt:75) includes `UNKNOWN` as a forward-compat fallback and the Now Playing UI has an UNKNOWN chip, but this `when` is exhaustive over the five current values with no `else`. If a newer server sends a state not in the current enum, the generated Moshi enum adapter throws while parsing the `Session` (surfacing as `Unexpected` via `execute()`), so the screen shows a generic error instead of degrading to UNKNOWN. Map unrecognized values to `PlaybackState.UNKNOWN` (and verify the generated adapter tolerates unknown enum values) so the intended graceful-degradation path is actually reachable. The 'unknown fields' test only covers unknown object fields, not unknown enum values.

app/src/main/java/io/github/xexanos/ratatoskr/ui/navigation/RatatoskrNavHost.kt
```diff
@@ -0,0 +1,133 @@
...
+                onOpenNowPlaying = { navController.navigate(Routes.NOW_PLAYING) },
+                onOpenSettings = { navController.navigate(Routes.SETTINGS) },
+            )
+        }
+
+        composable(
+            route = "${Routes.SPEAKERS}/{${Routes.ITEM_ID_ARG}}",
+            arguments = listOf(navArgument(Routes.ITEM_ID_ARG) { type = NavType.StringType }),
+        ) { backStackEntry ->
+            val itemId = backStackEntry.arguments?.getString(Routes.ITEM_ID_ARG).orEmpty()
```
Xexanos: Conventions + robustness: SPEC §13 records navigation as "Compose Navigation with a **sealed route graph**," but this uses stringly-typed routes (`object Routes { const val ... }` + string templates). CLAUDE.md requires structure decisions to go through the SPEC first, and this PR edits the SPEC — so either adopt the sealed/type-safe route graph or update §13. The concrete cost of the string approach shows up right here: `backStackEntry.arguments?.getString(ITEM_ID_ARG).orEmpty()` silently substitutes an empty id if the arg is ever missing, instead of failing loudly — exactly the class of bug the sealed graph was chosen to prevent. (Related: the token store lives in `network/persist/` while SPEC §13 records it under `auth/`; reconcile the SPEC and the package.)

app/build.gradle.kts
```diff
@@ -0,0 +1,62 @@
...
+        targetSdk = 36
+        versionCode = 1
+        versionName = "1.0"
+
+        testInstrumentationRunner = "androidx.test.runner.AndroidJUnitRunner"
+    }
+
+    buildTypes {
+        release {
+            optimization {
```
Xexanos, line 33: The `release` build type sets `optimization { enable = false }`, so R8/minification and resource shrinking are off for release. That makes `app/src/main/keepRules/rules.keep` dead — it's never exercised — and ships an unshrunk release. The risk is deferred: if optimization is later enabled, the Moshi/Retrofit reflection over the generated models can break because the keep rules were never validated against a real minified build. Either enable minification now (and validate the keep rules) or document why release ships unoptimized.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -0,0 +1,8 @@
+#Fri Jul 03 21:40:33 CEST 2026
+distributionBase=GRADLE_USER_HOME
+distributionPath=wrapper/dists
+distributionUrl=https\://services.gradle.org/distributions/gradle-9.4.1-bin.zip
```
Xexanos, line 4: F-Droid reproducibility: `distributionUrl` is set but there's no `distributionSha256Sum`, so the Gradle distribution is downloaded and executed without integrity verification (`validateDistributionUrl` only checks URL shape). For a reproducible/supply-chain-safe build, pin the checksum. Related: `gradle/gradle-daemon-jvm.properties` + the foojay-resolver convention in `settings.gradle.kts` introduce a build-time JDK toolchain download from `api.foojay.io`; on F-Droid's controlled/offline build environment that can fail the build or resolve a different JDK than the developer used, changing bytecode and breaking reproducibility. Consider pinning the wrapper checksum and provisioning the JDK without a network fetch.

Xexanos (commented):
Full review of the scaffold (8 subsystem finders across TLS, auth/token, API/mappers, UI/ViewModels, DI/lifecycle, build/security, conventions, tests; every candidate re-verified against the code myself). This is **in addition to** the 5 findings already posted (KeystoreCrypto decrypt crash, RatatoskrClient CancellationException, ConnectionManager race, NowPlaying slider, SignInScreen password) — those are not repeated here. 10 inline findings below, most-severe first.

Things I checked and deliberately did **not** file: the token DataStore backup exposure is real but the blob is Keystore-encrypted with a non-transferable key, so the consequence is the launch crash already filed, not plaintext leakage (still worth excluding it from backup, though). The missing explicit `https` scheme check is largely mitigated because `targetSdk 36` blocks cleartext by default (an `http://` URL fails rather than leaking) — worth an explicit guard for defense-in-depth. The `Example*Test` template tests should be deleted, and `RatatoskrClientTest` builds the client without the authenticator so the 401→refresh→retry path isn't covered end-to-end.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/persist/KeystoreCrypto.kt
Xexanos: Fixed in f934057: `decrypt` now returns `null` on a crypto failure (`AEADBadTagException`/other `GeneralSecurityException`) or malformed Base64, and no longer mints a key on the decrypt path (`existingKey()` is lookup-only). An unreadable token therefore leads to a re-login instead of a launch-time crash loop; `TokenStore`'s call sites already treat `null` as "no stored value".

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClient.kt
Xexanos: Fixed in f934057: `execute` and `executeUnit` now rethrow `CancellationException` before mapping, so a cancelled `viewModelScope` unwinds via structured concurrency instead of the dead coroutine writing a spurious error state. Added a test that a cancelled call propagates cancellation rather than returning a `Failure`.

app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
Xexanos: Fixed in f934057: `client()` now serialises construction behind a `Mutex` (double-checked, with `cached`/`cachedKey` marked `@Volatile`), so two concurrent callers can no longer each build an OkHttp stack and duplicate the single-flight refresh path. (The related "replaced clients are never shut down" finding is a separate medium item, not addressed in this batch.)

core-network/src/main/java/io/github/xexanos/ratatoskr/network/tls/PinnedTrustManager.kt
Xexanos: Addressed in d67b9ab as a documentation fix (the maintainer chose to keep the model). Platform-first is retained because SPEC section 6 mandates it (it covers the reverse-proxy-with-public-CA deployment). The overstated guarantee is what's corrected: the KDoc and SPEC section 6 now scope "rejected loudly on change" to the self-signed / local-CA path (where the platform chain fails, so the pin always decides), and document that with a publicly trusted certificate the pin is not consulted — trust is delegated to the CA. Strict leaf-pinning against public certs was considered and rejected because it would reject every routine renewal (e.g. Let's Encrypt) and force a manual re-trust each time.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/Mappers.kt
Xexanos: Implemented in PR #5, now merged into `scaffolding`. Resolved deliberately in the network layer rather than on the domain model: `RatatoskrClient` now adopts `rotatedTokens` from every session-returning call (`currentSession`/`startSession`/`pause`/`resume`/`seek`) and from the `stopSession` 200 body into the `TokenStore`, so the rotated pair is no longer dropped. `GenSession.toDomain()` and the domain `Session` are intentionally left unchanged — token handling stays inside `core-network` and the UI never sees `rotatedTokens`. The suppression side (`ConnectionManager.setSessionActive` → `TokenRefreshAuthenticator`) was already in place. Covered by four new tests in `RatatoskrClientTest` (adoption present/absent, and `stopSession` on both 200-with-pair and 204).

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
Xexanos: Fixed in 21df971. The slider no longer keys `remember` on the polled position. It tracks a `dragging` flag and a local `sliderPosition`, and a `LaunchedEffect(session.positionSeconds, dragging)` only follows the server value while `!dragging`. `onValueChange` sets `dragging = true`; `onValueChangeFinished` calls `onSeek(sliderPosition)` and then clears `dragging`. So a poll mid-drag can't reset the thumb, and the seek targets the dragged value.

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
Xexanos: Fixed in 21df971, both parts.

(1) Polling is no longer a self-running `viewModelScope` loop. `refresh()` is now public and the screen drives it via `lifecycleOwner.repeatOnLifecycle(Lifecycle.State.RESUMED) { while (isActive) { refresh(); delay(...) } }`, so a backgrounded app stops polling and stops holding the session active.

(2) `stop()` is no longer reverted by a late poll: `refresh()` returns early once `stopped` is set, and `applySession` now uses `copy(...)` instead of building a fresh `NowPlayingUiState`, so a poll can't clobber `stopped` or an unread error. Control actions (`pause`/`resume`/`seek`/`stop`) clear a stale error when they start, so it's the new action — not a later successful poll — that clears the previous message.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Fixed in 21df971. Search now flows through a `MutableStateFlow<String?>` collected once with `debounce { … } → distinctUntilChanged() → collectLatest { load(it) }`. `collectLatest` cancels an in-flight `load` when a newer query arrives, so results can no longer land out of order (last-typed wins), and the debounce collapses the per-keystroke burst. The initial (null) query debounces to 0 ms so the first library load stays immediate. `load` is now `suspend` and no longer launches its own coroutine per call.

app/src/main/java/io/github/xexanos/ratatoskr/MainActivity.kt
Xexanos: Fixed in 53712d6. `decideStartDestination` is now a `suspend` function run from a `LaunchedEffect` on `Dispatchers.IO`; `onCreate` no longer calls `runBlocking`. The NavHost renders once the route resolves, with a `CircularProgressIndicator` shown for the brief window while the DataStore reads and the Keystore decrypt run off the main thread.

app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
Xexanos: Fixed in 53712d6. `RatatoskrClient` now exposes `close()`, which the factory wires to shut down both the main and auth OkHttp clients' dispatcher executors and `evictAll()` their connection pools (mirroring `CertificateInspector`'s teardown). `ConnectionManager` calls it on the previous client when it is replaced (cache-miss for a new key) and in `invalidate()`, so a re-trust / server / fingerprint change no longer abandons a live thread pool and pooled sockets until GC.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/Mappers.kt
Xexanos: Fixed in 53712d6, with one caveat worth recording. A shared `ratatoskrMoshi()` now registers `EnumJsonAdapter.create(PlaybackState).withUnknownFallback(...)`, so an unrecognised state from a newer server deserializes instead of throwing and failing the whole `Session`. New test: `an unrecognised playback state falls back instead of failing the response`.

The deviation from the suggestion: I couldn't route the unknown value to the domain `UNKNOWN`. The kotlin generator ignores `enumUnknownDefaultCase` (I tried it — the generated enum still has only the five members), and the generated code can't be hand-edited, so there is no wire-level "unknown" member to fall back to. The fallback therefore lands on the neutral `PlaybackState.stopped` (it never implies audio is playing) rather than `UNKNOWN`; the domain `PlaybackState.UNKNOWN` stays as a defensive default. If routing to a real UNKNOWN chip matters, the clean route is a small hand-written Moshi adapter for the wire enum — happy to do that as a follow-up.

app/src/main/java/io/github/xexanos/ratatoskr/ui/auth/SignInScreen.kt
Xexanos: Fixed in 6f836a6. The password field now uses plain `remember`, so the plaintext is no longer written to the saved-instance-state Bundle (persisted to disk on process death). The username stays `rememberSaveable`.

app/build.gradle.kts
Xexanos: Documented in 6f836a6 (the "document why release ships unoptimized" option). R8 stays off for v1: the generated client uses reflective Moshi adapters (`@JsonClass(generateAdapter = false)`), so enabling shrinking safely needs Moshi/Retrofit keep rules validated against a real minified, on-device run — or, better, switching the generator to Moshi codegen adapters so no reflection is kept. Until that's done and verified, shipping unshrunk is safer than a release that only breaks at runtime. The build.gradle.kts release block now spells this out and it's tracked as a follow-up.

gradle/wrapper/gradle-wrapper.properties
Xexanos: Fixed in 6f836a6, both parts.

- `distributionSha256Sum` is now pinned in gradle-wrapper.properties (checksum from downloads.gradle.org for gradle-9.4.1-bin.zip), so the Gradle distribution is integrity-checked rather than executed unverified.
- Dropped the api.foojay.io JDK auto-download: removed the `toolchainUrl.*` entries from gradle-daemon-jvm.properties (keeping `toolchainVersion=21`) and the foojay-resolver-convention plugin from settings.gradle.kts. A build no longer fetches a JDK over the network — which broke offline/reproducible builds and could resolve a different JDK than the developer used. Gradle now runs the daemon on an already-installed JDK 21; verified a full compile + unit-test build still succeeds locally after stopping the old daemon.

app/src/main/java/io/github/xexanos/ratatoskr/ui/navigation/RatatoskrNavHost.kt
Xexanos: Fixed in c6a9c92.

Navigation is now a sealed, type-safe `Route` graph: each destination is a `@Serializable` member of `sealed interface Route`, `composable<Route.X>` replaces the string templates, and the speaker picker reads its id via `backStackEntry.toRoute<Route.Speakers>().itemId` — so the `.getString(...).orEmpty()` path that silently substituted an empty id is gone; a missing argument is now a type/serialization error, not a silent empty string. Added the kotlin-serialization plugin + kotlinx-serialization-core, and MainActivity's start destination is a `Route` rather than a string.

Side note (token-store location): aligned the SPEC to the code per the maintainer's call. SPEC section 13's module tree now records the encrypted token store and the trusted-server/fingerprint store under `core-network/persist/`, the bearer/refresh interceptors and single-flight guard under `api/`, and a `domain/` package for the domain models + ApiResult — matching the actual layout rather than the earlier `auth/` placeholder.

PR #4, opened 2026-07-06
UX design v1, knot spinner, and UX guardrails

Xexanos:
## What this adds

Foundational UX work on top of `scaffolding`, all additive (the polished screens already on `scaffolding` are untouched):

- **UX design v1** — `docs/ux-design.html`: a self-contained design doc with principles, the user flow, brand-derived design tokens, mockups of all six screens, states, and an implementation/localization plan.
- **Animated knot spinner** — the logo generator now emits `docs/logo/ratatoskr-spinner.svg` (`spinnerSvg()` / `centerlinePath()` in `tools/logo`): the real woven knot as a static background with a copper “squirrel” + comet trail running the strand’s own centreline, embedded via SMIL. Deterministic (`gradlew :tools:logo:run` leaves the other logo SVGs byte-identical). The app will redraw it in Compose (VectorDrawable has no SMIL).
- **UX review tooling** — `.claude/skills/ux-review` (single source of truth for what “UX-compliant” means) + `.claude/agents/ux-review.md` (read-only reviewer that loads the skill).
- **Deterministic CI floor** — `scripts/check-ux.sh` + `.github/workflows/ci.yml` (gate + `lintDebug`/`test`/`assembleDebug`, contract submodule checked out).
- **NowPlaying** — cover shape uses `MaterialTheme.shapes.extraLarge` (token, not a literal) and its copy is in `strings.xml`.

## Gate posture (please read)

The gate is introduced onto an existing codebase, so it enforces what the code already satisfies and reports the rest — the standard way to adopt a linter:

| Rule | Status |
|---|---|
| Hardcoded colors outside `theme/` | **fatal** (clean) |
| Real Material-2 component imports | **fatal** (clean; `material.icons.*` correctly allowed) |
| `RoundedCornerShape(n.dp)` corner radii | advisory (scaffolding uses custom radii like `20.dp` with no M3 token) |
| Hardcoded UI strings | advisory (scaffolding screens not yet in `strings.xml`) |

Flip `STRICT_SHAPES` / `STRICT_STRINGS` to fatal per rule as the screens migrate.

## Note on the rebase

`scaffolding` independently polished the five screens (cards, icons, empty states). That superseded the screen-level string extraction from this branch, so I resolved the rebase in favour of scaffolding’s screens and trimmed `strings.xml` to what’s actually used (NowPlaying). Re-extracting those screens’ copy into `strings.xml` and migrating their radii to `MaterialTheme.shapes` are the natural follow-ups (then the two advisory rules go fatal).

## Not included / follow-up

- Screen redesigns from the design doc still to build on top of scaffolding’s polish (Library mini-player, NowPlaying ±10 s + “PLAYING ON”, Speakers resume-context card, etc.).
- Localizing ViewModel/`ErrorText.kt` error strings (needs a small Context/architecture change).
- Screenshot golden tests (Roborazzi) once the screens are final.


.claude/agents/ux-review.md
```diff
@@ -0,0 +1,25 @@
+---
+name: ux-review
+description: Reviews Ratatoskr Android UI changes for drift against the checked-in UX design and Material 3, in isolated context. Use PROACTIVELY before merging UI work, or when asked whether a diff matches the design. Read-only; returns ranked findings, does not edit.
+tools: Read, Grep, Glob, Bash
+model: inherit
```
Xexanos: Hier sollten wir einmal schauen, ob ein konkretes Modell mehr Sinn ergibt.

scripts/check-ux.sh
```diff
@@ -0,0 +1,65 @@
...
+  else echo "  warn: move UI copy into strings.xml (English source), then flip STRICT_STRINGS=1."; fi
+else echo "  ok"; fi
+
+echo "== 5. Emoji used as UI glyphs (use Material icons) [$( [ "$STRICT_EMOJI" = 1 ] && echo FATAL || echo advisory )] =="
+# Fixed-string (byte-wise) match on a denylist of icon-like glyphs — locale-independent, so
+# it behaves the same on git-bash and the CI runner, and never touches umlauts/em-dash/quotes.
+emoji_hits="$(grep -rnF --include='*.kt' \
+  -e '▶' -e '⏸' -e '⏹' -e '⏭' -e '⏮' -e '⏩' -e '⏪' -e '♪' -e '♫' -e '♬' \
+  -e '🔀' -e '🔁' -e '🔊' -e '🔈' -e '🔇' -e '🎵' -e '🎶' -e '🎧' -e '📚' -e '📖' \
+  -e '🔍' -e '✅' -e '❌' -e '⚠' -e '➕' -e '➖' -e '🔒' -e '🔑' -e '⚙' -e '🏠' -e '⭐' "$UI" || true)"
```
Xexanos: Is it possible to simply check for **all** emojis? Such an blocklist is prone to errors. If we have some emojis, that are allowed to be used, we should use an explicit allowlist.

scripts/check-ux.sh
```diff
@@ -0,0 +1,65 @@
...
+# drift that never needs a human eye. Run locally with `bash scripts/check-ux.sh`; CI runs
+# it on every push/PR.
+#
+# Posture: this gate was introduced onto an existing codebase, so it enforces only what the
+# code already satisfies and *reports* the rest, exactly as you'd adopt any linter:
+#   - colors and real Material-2 component imports are FATAL (the code is already clean);
+#   - corner radii, hardcoded strings, and emoji-as-icons are ADVISORY for now — the screens
+#     predate these rules (custom radii like 20.dp have no M3 shape token; copy is not yet in
+#     strings.xml; some empty states use emoji). Flip STRICT_SHAPES / STRICT_STRINGS /
+#     STRICT_EMOJI to 1 per rule as the screens are migrated.
```
Xexanos: I would appreciate it, when the Errors are fixed with this PR, so we can activate the Rules. Since the codebase is small this should not be that much work.

.claude/agents/ux-review.md
Xexanos: Done in f5a43e3 — pinned to `model: opus`. For an on-demand adversarial reviewer the value is judgment, and it isn't run on every commit, so I favoured depth. If you'd rather trade some depth for speed/cost on frequent runs, `sonnet` is a one-line change.

scripts/check-ux.sh
Xexanos: Agreed — the denylist was fragile (it had already missed `⧉` and `◉` in SpeakersScreen). Replaced it in f5a43e3 with `scripts/lint_emoji.py`: a range-based detector that flags **all** emoji/pictograph blocks, with an explicit `ALLOWLIST` (empty today) for anything we ever want to permit. Locale-independent + UTF-8 output so a non-UTF-8 console can't crash it and silently pass.

scripts/check-ux.sh
Xexanos: Agreed on activating the rules. Status:

- **Strings** and **corner radii** are unblocked and mechanical — I'll move all screen copy into `strings.xml` and all `RoundedCornerShape(n.dp)` onto a shape scale defined in the theme (preserving the current 12/16/20 radii), then flip `STRICT_STRINGS` and `STRICT_SHAPES` to fatal.
- **Emoji** has one real snag worth a decision: the app only depends on `material-icons-core`, which has `PlayArrow`/`Search`/`Settings` but **not** `Pause`, `Stop`, `Speaker`, `VolumeOff`, `MusicNote`, `Replay10`/`Forward10`, `SkipNext`/`SkipPrevious` — those live in `material-icons-extended`. Activating the emoji rule therefore means either **(a)** hand-authoring a few small local `ImageVector`s for the missing glyphs (keeps the dependency set small — F-Droid-friendly, my recommendation), or **(b)** adding `material-icons-extended` (much larger artifact). I'll take (a) unless you prefer (b), then flip `STRICT_EMOJI` to fatal.

Xexanos: Done in 8a2c180 — went with `material-icons-extended` per your pick. All six screens now use `strings.xml`, `MaterialTheme.shapes` (scale defined once in the theme, 12/16/20 dp preserved), and Material icons (▶⏸⏹ ♪ 🔇 ⧉ ◉ 📚 🔍 → icons, each with a `contentDescription` — this also closes the NowPlaying transport TalkBack gap). **All five gate rules are now fatal.** Verified locally: `lintDebug` + `test` + `assembleDebug` pass and the gate is green with every rule strict.

scripts/check-ux.sh
Xexanos: Refined further in 3aa8ce8 per your point about gaps: the block list did have holes (box-drawing, newer emoji blocks), so it's now a **single threshold** — reject every character from U+2190 up — with a small explicit **allowlist** for the non-emoji symbols that legitimately appear in source (`→ ← ↑ ↓ ≤ ≥ ≠`). Gap-free for emoji, and the cut sits just above the license-header `©` / `™` / typographic marks so those don't false-positive. Verified: box-drawing `─` and `◉` now flag; `→` and `café/Straße` don't.

Xexanos: Taken further per your ASCII request (410dafe): instead of an emoji-specific check, gate rule 5 now enforces **ASCII-only source** — `lint_ascii.py` flags any char > U+007F, which subsumes emoji and closes the range gaps entirely. Scope: Kotlin + Gradle/shell/python + prose docs + the base `strings.xml`; exempt: the rendered design doc, SVGs, translations (`values-*/`, fastlane locales), and vendored `gradlew`. Existing non-ASCII across code and docs was converted (`->`, `--`, `...`, ASCII tree diagrams, etc.).

.gitattributes
```diff
@@ -18,7 +18,7 @@ gradlew            text eol=lf
 *.yaml             text eol=lf
 *.yml              text eol=lf
 
-# Binary assets — never touch.
+# Binary assets -- never touch.
```
Xexanos:
```suggestion
# Binary assets - never touch.
```

.gitignore
```diff
@@ -22,9 +22,16 @@ keystore.properties
 *.iml
 .DS_Store
 
-# Installed agent/skills (local tooling, not project code)
-skills-lock.json
-.claude/skills/
+# Installed skills (third-party local tooling, not project code). Our own project
+# skills and agents ARE committed -- installed skills are re-excluded per entry below.
```
Xexanos:
```suggestion
# skills and agents ARE committed - installed skills are re-excluded per entry below.
```

.gitignore
```diff
@@ -35,7 +42,7 @@ tools/logo/.fontcache/
 *.log
 
 # Generated API client (generated at build time from the OpenAPI contract; never
-# committed — see SPEC section 4). AGP writes it under core-network/build
+# committed -- see SPEC section 4). AGP writes it under core-network/build
```
Xexanos:
```suggestion
# committed - see SPEC section 4). AGP writes it under core-network/build
```

Xexanos (commented):
Please remove double dasches

PR #19, opened 2026-07-12
Enable R8 release shrinking via Moshi codegen adapters (SPEC section 8)

Xexanos:
Takes up the next task the SPEC itself sequences ("right after the integration-test layer", section 8): release shrinking. All section-9 test layers are in place, so the precondition is met.

## What changed

1. **Moshi codegen instead of reflection** — `moshiCodeGen=true` on the openapi generator, `moshi-kotlin-codegen` via KSP in `core-network`, Moshi 1.15.2 (first codegen release supporting KSP2, the only mode AGP 9 runs), `moshi-kotlin` (and with it `kotlin-reflect`) removed. The generated sources are now registered as *kotlin* sources so the KSP task is guaranteed to see them. Enums are untouched by codegen, so the unknown-`PlaybackState` fallback keeps working; `MappersTest` gains the strict-side parity case (missing required field still throws).
2. **R8 on for release** — `optimization { enable = true }`. Keep rules are minimal: `-dontobfuscate` (F-Droid users never get a mapping file; also simplifies reproducible-build verification), plus explicit keeps for the reflectively looked-up codegen adapters as insurance in case the codegen-generated rules don't propagate from the library module.
3. **Testable `minified` build type** — the release R8 configuration, debug-signed and debuggable so instrumentation can attach; selected via `-PminifiedTests` so per-PR debug test jobs are untouched. Build-type-scoped keep rules preserve the test-only entry points (`AppContainer.reset`, the factory the smoke test drives). New `MinifiedWireSmokeTest` mirrors the component layer's wire checks inside the app process — the component suite lives in `core-network` and can never run against shrunk output.
4. **Post-merge validation workflow** (`release-validation.yml`, per the agreed cadence): on every push to `main`, run `AppFlowTest` + `MinifiedWireSmokeTest` against the minified build on an API 36 emulator; on success publish the tested APKs to the rolling `latest-main` pre-release (debug-signed minified for the black-box E2E harness, unsigned release for F-Droid), then a documented placeholder for the `ratatoskr-e2e` dispatch. Per-PR, `ci.yml` now also runs `:app:assembleRelease` so a broken keep rule fails cheaply on the JVM.
5. **SPEC sections 8/9 and `docs/testing.md`** updated to record the landed decisions.

## Verification

The dev environment for this change has no access to Maven Central/Google's Maven (egress policy), so builds could not be run locally — CI on this PR is the verification: lint + unit tests (deserialization parity via `MappersTest`), `assembleDebug`/`assembleRelease` (R8 config), and the instrumented component + integration suites (converter parity, TLS, auth). The on-device minified run happens post-merge via the new workflow; it can also be exercised beforehand with `workflow_dispatch` on this branch if desired.

Known follow-ups if CI flags them (will fix on this PR): AGP 9 new-DSL property naming in the `minified` block, and KSP visibility of the generated sources (fallback documented in the build script comment).

Before/after release APK sizes will be posted here once `assembleRelease` has run in CI.

https://claude.ai/code/session_01Sq4kXnpQ8zpk2VZAFPyUSN

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sq4kXnpQ8zpk2VZAFPyUSN)_


.github/workflows/release-validation.yml
```diff
@@ -0,0 +1,150 @@
...
+          mkdir -p publish
+          cp app/build/outputs/apk/minified/app-minified.apk publish/ratatoskr-minified.apk
+          cp app/build/outputs/apk/release/app-release-unsigned.apk publish/ratatoskr-release-unsigned.apk
+
+      # Rolling pre-release: the tag and release body always point at the latest validated
+      # main commit; assets are replaced on every run.
+      - name: Publish the validated APKs to the latest-main pre-release
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: latest-main
```
Xexanos: latest implies (at least to me) that this is something stable. I think something like `testing` or `testing-<hash>` would be more fitting.

.github/workflows/release-validation.yml
```diff
@@ -0,0 +1,150 @@
...
+          java-version: 21
+
+      - name: Set up Gradle
+        uses: gradle/actions/setup-gradle@v4
+
+      # Same key as instrumented-tests.yml. PR-branch caches are not visible to main, so
+      # this hits only once a warm run has happened on main; on a miss the runner creates
+      # the AVD itself (slower, acceptable post-merge).
+      - name: Restore the warmed AVD snapshot
+        uses: actions/cache@v4
```
Xexanos: **Full `actions/cache@v4` here permanently poisons the shared AVD key with a cold, snapshot-less AVD — slowing every future PR's API-36 jobs.**

Mechanics: `instrumented-tests.yml` runs on `pull_request`, so its warm caches live in PR-scoped entries — main cannot see them, and nothing ever warms this key **on main** (the comment's "once a warm run has happened on main" never happens). So the restore misses on the first main run; the emulator boots with `-no-snapshot-save` (no quickboot snapshot written); then this action's **post step saves** that snapshot-less AVD under `avd-api36-x86_64-default-pixel_6` in the **default-branch scope**, which every PR branch can read.

Consequence: every future PR's `warm-avd` job gets an exact-key cross-scope hit → `cache-hit == true` → warming is skipped → all four API-36 test jobs cold-boot (3–6 min each), forever — cache keys are immutable, so the warm job never gets another chance to save. `instrumented-tests.yml` gets away with full `actions/cache` in its test jobs only because `warm-avd` guarantees the key is already saved warm in-scope first; this workflow has no such guarantee.

Fix: use `actions/cache/restore@v4` (restore-only) here — which also makes the comment true.

.github/workflows/release-validation.yml
Xexanos: **The `latest-main` git tag never moves after the first run.**

On run 1 the action creates the release and the tag at that commit. On run 2+ it finds the existing release by tag and PATCHes it (name, body, assets) — it never updates the tag ref (long-standing behavior, softprops/action-gh-release#171, and there is no input that moves an existing tag). Result: the body says `Commit: <new sha>` and the assets are fresh, but `git checkout latest-main`, GitHub's tag/compare view, and any tooling resolving the tag get the first-ever main commit, forever.

Fix: force-move the tag in a step before publishing —

```yaml
- name: Move the latest-main tag to this commit
  run: |
    git tag -f latest-main ${{ github.sha }}
    git push --force origin refs/tags/latest-main
```

(Avoid delete-and-recreate of the *release* — that briefly 404s the stable download URLs and re-notifies watchers.)

.github/workflows/ci.yml
```diff
@@ -67,4 +67,7 @@ jobs:
         run: bash scripts/check-ux.sh
 
       - name: Lint, assemble debug and run unit tests
-        run: ./gradlew --no-daemon lintDebug testDebugUnitTest assembleDebug
+        # assembleRelease runs R8 (SPEC section 8) on the JVM, so a broken keep rule or
+        # shrinker config fails the PR cheaply here; the on-device validation of the shrunk
+        # build runs post-merge (release-validation.yml).
+        run: ./gradlew --no-daemon lintDebug testDebugUnitTest assembleDebug :app:assembleRelease
```
Xexanos: **The per-PR gate never runs R8 for the `minified` build type — `src/minified/keepRules/tests.keep` is first parsed post-merge.**

`assembleRelease` runs R8 with `src/main/keepRules` only; the debug connected jobs don't shrink at all. So a typo'd rule in `tests.keep` — or a keep rule that stops matching after a refactor of `AppContainer`/`RatatoskrClientFactory` — sails through PR CI green and only fails in `release-validation.yml` on main, which then also blocks publishing `latest-main` (and, later, the E2E trigger). The comment above ("a broken keep rule or shrinker config fails the PR cheaply here") doesn't hold for minified-only config.

Cheap fix, same JVM-only cost class: add `:app:assembleMinified` to this line — it runs the minified R8 pass (parsing `tests.keep`) without an emulator.

.github/workflows/release-validation.yml
```diff
@@ -0,0 +1,150 @@
...
+  workflow_dispatch:
+
+permissions:
+  # Needed to create/update the rolling `latest-main` pre-release and its tag.
+  contents: write
+
+concurrency:
+  # A newer main push supersedes an in-flight validation of an older commit.
+  group: release-validation-${{ github.ref_name }}
+  cancel-in-progress: true
```
Xexanos, line 50: **`cancel-in-progress` on a job that publishes can strand mixed-generation assets on `latest-main`.**

The publish action patches the release body (→ new sha) first, then per file deletes the same-name asset and uploads its replacement — not atomic across the two APKs. A newer main push cancels the in-flight run wherever it happens to be; landing between the two uploads leaves a new `ratatoskr-minified.apk` next to an old `ratatoskr-release-unsigned.apk`, with the body already claiming the new sha, until the superseding run finishes (~25 min). The window is small, but the payload is the F-Droid-bound unsigned APK — from the wrong commit.

Mitigation: split into two jobs — `validate` (keeps `cancel-in-progress: true`) and `publish` (`needs: validate`, its own concurrency group with `cancel-in-progress: false`) — so cancellation can only ever hit before publishing starts.

Related, same block: `workflow_dispatch` (line 29) from a non-main ref uses a per-ref concurrency group, escapes superseding entirely, and would overwrite the `latest-main` assets with branch-built APKs. Guard the publish step with `if: github.ref == 'refs/heads/main'`.

.github/workflows/release-validation.yml
```diff
@@ -0,0 +1,150 @@
...
+
+jobs:
+  validate-and-publish:
+    name: Validate the shrunk build & publish the pre-release
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
```
Xexanos: **30 minutes is tight for what is — on main — permanently the cold path.**

Nothing ever warms the AVD snapshot on main (see the cache comment above), so every run of this job pays: AVD creation + cold boot (`instrumented-tests.yml` budgets 20 min for its `warm-avd` job alone) **plus** a full R8 minified build (slower than debug) + two instrumented suites + a second R8 pass (`assembleRelease`) + publish. `instrumented-tests` budgets 15 min for a test job that starts from a *warm* snapshot and builds plain debug.

Suggest 45 min — or add a small AVD-warm step on main so the snapshot restore actually hits, which fixes the cause rather than the budget.

app/src/main/keepRules/rules.keep
```diff
@@ -1,12 +1,22 @@
...
+# also easier to verify for the reproducible build and keeps the minified
+# instrumented run free of test-APK mapping fragility. Costs a modest part of the
+# size win - the bulk comes from shrinking, which stays fully active.
+-dontobfuscate
+
+# Moshi codegen adapters are looked up reflectively (Class.forName(model + "JsonAdapter")),
+# which the shrinker cannot trace. moshi-kotlin-codegen generates equivalent per-model rules
+# inside core-network (META-INF/proguard); these two are deliberate insurance in case those
+# resource-borne rules do not propagate from the library module into this module's R8 run.
+-keepnames @com.squareup.moshi.JsonClass class io.github.xexanos.ratatoskr.network.generated.model.**
```
Xexanos: **This `-keepnames` rule is a no-op — it provides none of the stated insurance.**

`-keepnames` is `-keep,allowshrinking`: it only prevents *renaming*, which `-dontobfuscate` (line 13) already disables globally, while leaving the classes fully strippable. So as written it protects nothing. The real insurance is the adapter `-keep` below; if the model classes themselves are meant to be insured too, this needs to be a plain `-keep` — otherwise it's clearer to delete the line.

While touching the adapter rule below: it keeps only `<init>(Moshi)`. All 16 current generated models are non-generic, so that's complete today — but a future parameterized model gets a `<init>(Moshi, Type[])` constructor this rule doesn't keep, failing exactly in the propagation-failure scenario the insurance exists for. One extra member line makes it future-proof:

```
    <init>(com.squareup.moshi.Moshi, java.lang.reflect.Type[]);
```

.github/workflows/release-validation.yml
Xexanos: Resolved by the switch to per-commit `testing-<short-sha>` releases rather than a force-move step: each run creates a *new* tag at `github.sha` (with `target_commitish: ${{ github.sha }}`), so a tag never has to move and always resolves to the commit it names. No stale-tag class of bug to guard against.

---
_Generated by [Claude Code](https://claude.ai/code)_

.github/workflows/release-validation.yml
Xexanos: The mixed-generation stranding is gone with per-commit releases: each run publishes to its own `testing-<sha>` release, so nothing is shared across commits. A cancel-in-progress hit can at worst leave one *superseded* commit's release half-populated (pruned later); the newest commit's run always publishes a complete release, and the E2E harness targets that. So I kept it a single job rather than splitting validate/publish.

The `workflow_dispatch`-from-a-branch half I did take: publish and the E2E trigger are now guarded with `if: github.ref == 'refs/heads/main'`.

---
_Generated by [Claude Code](https://claude.ai/code)_

docs/testing.md
```diff
@@ -80,7 +93,10 @@ The strategy above is the target. Current state:
   the black-box E2E harness. The suites run as parallel per-suite CI jobs; the
   **component** suite runs on **both API 26 (minSdk) and API 36 (target)** — the
   layer whose behaviour diverges by API level (Conscrypt/TLS, Keystore) — while
-  the whole-app UI and accessibility suites run on **API 36 only**.
+  the whole-app UI and accessibility suites run on **API 36 only**; the shrunk
+  build (`minified`, SPEC section 8) is validated post-merge on `main` by
+  `release-validation.yml` (AppFlowTest + MinifiedWireSmokeTest on API 36), which
+  then publishes the tested APKs to the `latest-main` pre-release.
```
Xexanos: The tag in this description ist not correct.

.github/workflows/release-validation.yml
Xexanos: Done: the rolling \`latest-main\` release is replaced by per-commit \`testing-<short-sha>\` pre-releases (8cb8159), plus a scheduled cleanup job that prunes \`testing-*\` older than 14 days (keeping the most recent 3 as a floor).

.github/workflows/release-validation.yml
Xexanos: Fixed in 8cb8159: switched to \`actions/cache/restore@v4\` (restore-only) so this workflow can never *save* a cold, snapshot-less AVD under the shared default-branch key that the PR \`warm-avd\` jobs read.

.github/workflows/ci.yml
Xexanos: Fixed in 8cb8159: added \`:app:assembleMinified\` to the gradle line, so \`src/minified/keepRules/tests.keep\` gets an R8 pass per PR on the JVM instead of first post-merge.

.github/workflows/release-validation.yml
Xexanos: Bumped to 45m in 8cb8159, with a comment noting this is permanently the cold path on main. I kept restore-only (per the cache thread) rather than warming/poisoning the shared key, so the budget is the fix here.

app/src/main/keepRules/rules.keep
Xexanos: Fixed in 8cb8159: dropped the no-op \`-keepnames\` (the models are reachable by reference, and the adapter keep below references them too), kept the adapter \`-keep\`, and added the \`<init>(com.squareup.moshi.Moshi, java.lang.reflect.Type[])\` ctor so a future parameterized model stays covered.

docs/testing.md
Xexanos: Fixed: all \`latest-main\` references updated to per-commit \`testing-<short-sha>\` in docs/testing.md and SPEC sections 8/9 (64f948a). The measured APK sizes are also recorded in SPEC section 8 now (aa18037).

PR #32, opened 2026-07-13
Fix core-network instrumented tests under the okhttp 4.12.0 → 5.4.0 bump

Xexanos:
## Context

This branch carries the okhttp `4.12.0 → 5.4.0` bump from #29 plus the fix that gets its CI green. On the bump alone, the `:core-network:connectedDebugAndroidTest` job [crashed on the first HTTPS request](https://github.com/Xexanos/ratatoskr-app/actions/runs/29263882493/job/86864108710?pr=29):

```
FactoryAuthComponentTest > theBearerTokenIsAttachedToRequests FAILED
  java.lang.NullPointerException: getRequestedServerNames(...) must not be null
    at okhttp3.internal.platform.Platform.getHandshakeServerNames(Platform.kt:135)
    at mockwebserver3.internal.MockWebServerSocket.getHandshakeServerNames(MockWebServerSocket.kt:95)
    at mockwebserver3.internal.RecordedRequestFactoryKt.RecordedRequest(RecordedRequestFactory.kt:51)
...
Test run failed to complete. Instrumentation run failed due to Process crashed.
```

The NPE is thrown on a MockWebServer background thread, so it crashes the whole instrumentation process — only 1 of 26 tests ran before the run aborted.

## Root cause

okhttp 5 routes `okhttp3.mockwebserver` through `mockwebserver3`, whose `RecordedRequest` now captures the server-side TLS SNI names (`handshakeServerNames`) for **every** request. On Android, `Platform.getHandshakeServerNames` reads `ExtendedSSLSession.getRequestedServerNames()`, and Conscrypt returns `null` there when the client sent no SNI extension — which the Kotlin non-null check turns into the crash above. okhttp 4's mockwebserver never read this, so the tests passed before the bump.

The client only emits an SNI extension for non-IP hosts, but `HttpsMockServer` connected via `MockWebServer.url()`, whose host resolves to a loopback **IP literal** on the emulator → no SNI → the null.

## Change

`HttpsMockServer` now forces both the base URL host and the served certificate's SAN to the DNS name `localhost`, so the handshake carries SNI and hostname verification still passes. `localhost` resolves to the loopback MockWebServer binds to, so connectivity is unchanged. One test-fixture file, no production/client code touched.

## Verification

No Android SDK/emulator in this environment, so the fix relies on the PR's `connectedDebugAndroidTest` CI job for end-to-end confirmation.

---
_Generated by [Claude Code](https://claude.ai/code/session_01AjLpkqAYvfNpydQkBqoJRC)_


Xexanos:
Deferring the okhttp 5.4.0 upgrade for now.

The API-26 instrumented-test crash is a known, still-open okhttp bug — [square/okhttp#9523](https://github.com/square/okhttp/issues/9523): on Android ≤ 28 emulators Conscrypt returns `null` from `ExtendedSSLSession.getRequestedServerNames`, and mockwebserver3 records that per request without handling the null, which crashes the whole instrumentation process. It is not fixed in 5.4.0, and a one-line upstream fix is pending.

Getting the suite green required an `SSLSocket`-factory wrapper that reimplements ~50 delegation methods just to hide the `ExtendedSSLSession` from that read — disproportionate framework-mirroring for a bug that upstream will fix (their own suggested workaround is simply to skip the affected SDKs). Rather than carry that, we'll stay on okhttp 4.12.0 and revisit once #9523 lands.

Closing this PR; the branch is preserved so we can resume from the bump when the upstream fix ships.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #35, opened 2026-07-13
fix(tls): ignore certificate hostname on the pinned connection

Xexanos:
## Problem

The pinned **production** OkHttp client (`RatatoskrClientFactory`) kept OkHttp's **default hostname verifier** on every path. After a server is trusted by fingerprint, each request still verified the certificate hostname against its SAN - so the app could only reach the server at an address listed in the cert. That **breaks the normal case of connecting by LAN IP** (`https://192.168.x.x:8080`) and the E2E case (emulator -> `10.0.2.2`).

Surfaced while reviewing the server's zero-config self-signed TLS (ratatoskr-server#42): that cert's SAN is `localhost/127.0.0.1/ratatoskr`, so a phone connecting by LAN IP would fail hostname verification after trusting it.

## Fix

**Pin-aware hostname verification** (`PinnedHostnameVerifier`): the hostname check is skipped only when the served leaf certificate matches the user-confirmed fingerprint - the user confirmed *that exact certificate*, so it is trustworthy at any address. Everything else stays host-bound through OkHttp's standard verifier.

That distinction matters because `PinnedTrustManager` is **not** exact-cert pinning: it tries platform trust first and consults the pin only on its failure (SPEC section 6, reverse-proxy trade-off). An earlier revision of this PR disabled hostname verification wholesale, which on the platform-trusted path would have accepted a validly issued certificate for *any* domain - caught in review; see the resolved threads.

- `RatatoskrClientFactory`: `.hostnameVerifier(PinnedHostnameVerifier(fingerprint))` on `baseBuilder` (both the main and refresh clients inherit it).
- `PinnedHostnameVerifierTest` (JVM): covers the platform path the instrumented suite cannot reach - a non-matching leaf on a non-matching hostname is rejected (the attack case), no-pin clients stay host-bound, delegation to the default verifier works, unverified peers fail closed.
- `HttpsMockServer`: the served cert's SAN deliberately does not match the host it serves on, so the instrumented HTTPS suite proves hostname-independence **of the pin path** through the real Conscrypt stack; `FactoryTlsPinComponentTest` names the behaviour explicitly.
- SPEC section 6 records the pin-aware hostname rule.

## Verification

- `:core-network:testDebugUnitTest` - green, including the 5 new verifier tests.
- `:core-network:connectedDebugAndroidTest` on an emulator (API 36) - all 27 tests green, including the foreign-SAN pin suite (which also proves the verifier can read the peer certificates from Conscrypt's `SSLSession`).

Prerequisite for the cross-component E2E suite. ratatoskr-server#42 needs no change once this lands (its self-signed cert then works for any client address).

### DIFF
--- a/core-network/src/androidTest/java/io/github/xexanos/ratatoskr/network/component/FactoryTlsPinComponentTest.kt
+++ b/core-network/src/androidTest/java/io/github/xexanos/ratatoskr/network/component/FactoryTlsPinComponentTest.kt
@@ -28,6 +28,9 @@ import org.junit.runner.RunWith
  * skipped) is deliberately NOT covered here: the trust manager consults only the system trust
  * store, which a test-minted CA is not in, so a self-signed MockWebServer cannot reach that path.
  * It stays a documented gap (SPEC section 9).
+ *
+ * The fixture's certificate SAN deliberately does not match the host it serves on, so these
+ * tests also prove the pinned client ignores the certificate hostname (SPEC sections 6 and 14).
  */
 @RunWith(AndroidJUnit4::class)
 class FactoryTlsPinComponentTest {
@@ -45,6 +48,20 @@ class FactoryTlsPinComponentTest {
 
         assertTrue("expected Success, was $result", result is ApiResult.Success)
         assertTrue((result as ApiResult.Success).data.isEmpty())
+    }
+
+    @Test
+    fun aPinnedConnectionIgnoresTheCertificateHostname() = runBlocking {
+        // HttpsMockServer serves a certificate whose SAN never matches the loopback host it runs
+        // on, yet the connection succeeds: exact-cert pinning supersedes hostname verification,
+        // which the factory disables (SPEC sections 6 and 14). Guards against re-enabling the
+        // default hostname verifier on the pinned client - which would break connecting to a
+        // server by a LAN IP (and the E2E emulator's 10.0.2.2).
+        https.enqueueJson("[]")
+
+        val result = client(https.fingerprint).listSpeakers()
+
+        assertTrue("expected Success despite a non-matching cert SAN, was $result", result is ApiResult.Success)
     }
 
     @Test
--- a/core-network/src/androidTest/java/io/github/xexanos/ratatoskr/network/component/FactoryTlsPinComponentTest.kt
+++ b/core-network/src/androidTest/java/io/github/xexanos/ratatoskr/network/component/FactoryTlsPinComponentTest.kt
@@ -30,7 +30,9 @@ import org.junit.runner.RunWith
  * It stays a documented gap (SPEC section 9).
  *
  * The fixture's certificate SAN deliberately does not match the host it serves on, so these
- * tests also prove the pinned client ignores the certificate hostname (SPEC sections 6 and 14).
+ * tests also prove the PIN path ignores the certificate hostname (SPEC section 6). The
+ * platform-trusted path stays host-bound through the pin-aware verifier; being unreachable
+ * here (same reason as above), that binding is covered by PinnedHostnameVerifierTest on the JVM.
  */
 @RunWith(AndroidJUnit4::class)
 class FactoryTlsPinComponentTest {
@@ -53,10 +55,11 @@ class FactoryTlsPinComponentTest {
     @Test
     fun aPinnedConnectionIgnoresTheCertificateHostname() = runBlocking {
         // HttpsMockServer serves a certificate whose SAN never matches the loopback host it runs
-        // on, yet the connection succeeds: exact-cert pinning supersedes hostname verification,
-        // which the factory disables (SPEC sections 6 and 14). Guards against re-enabling the
-        // default hostname verifier on the pinned client - which would break connecting to a
-        // server by a LAN IP (and the E2E emulator's 10.0.2.2).
+        // on, yet the connection succeeds: the confirmed pin carries the connection, so the
+        // pin-aware hostname verifier ignores the hostname (SPEC section 6). Guards against
+        // losing that - a hostname-bound pin path would break connecting to a server by a LAN IP
+        // (and the E2E emulator's 10.0.2.2). Also proves, through the real Conscrypt stack, that
+        // the verifier can read the peer certificates it needs for the pin comparison.
         https.enqueueJson("[]")
 
         val result = client(https.fingerprint).listSpeakers()
--- a/core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientFactory.kt
+++ b/core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientFactory.kt
@@ -67,6 +67,12 @@ object RatatoskrClientFactory {
 
         val baseBuilder = OkHttpClient.Builder()
             .sslSocketFactory(sslSocketFactory, trustManager)
+            // Exact-certificate trust-on-first-use pinning ([PinnedTrustManager]) already
+            // guarantees the server presents the one certificate the user confirmed by
+            // fingerprint, so hostname verification is redundant here - and enforcing it would
+            // break connecting to a server at an address not in its cert (a LAN IP, etc.). Let
+            // the pin alone decide (SPEC sections 6 and 14); mirrors CertificateInspector.
+            .hostnameVerifier { _, _ -> true }
             .connectTimeout(15, TimeUnit.SECONDS)
             .readTimeout(30, TimeUnit.SECONDS)
             .addInterceptor(logging)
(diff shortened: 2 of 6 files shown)


core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientFactory.kt
```diff
@@ -67,6 +67,12 @@ object RatatoskrClientFactory {
 
         val baseBuilder = OkHttpClient.Builder()
             .sslSocketFactory(sslSocketFactory, trustManager)
+            // Exact-certificate trust-on-first-use pinning ([PinnedTrustManager]) already
+            // guarantees the server presents the one certificate the user confirmed by
+            // fingerprint, so hostname verification is redundant here - and enforcing it would
+            // break connecting to a server at an address not in its cert (a LAN IP, etc.). Let
+            // the pin alone decide (SPEC sections 6 and 14); mirrors CertificateInspector.
+            .hostnameVerifier { _, _ -> true }
```
Xexanos: **Critical:** This comment's premise is inverted relative to what `PinnedTrustManager` actually does. It is not exact-cert pinning — it tries **platform trust first** and only falls back to the pin when the chain fails (SPEC §6, reverse-proxy trade-off). `checkServerTrusted` never checks the hostname; with this verifier disabled, nothing does on the platform path.

Attack: an on-path attacker (rogue AP, ARP spoofing) presents a valid publicly-trusted certificate for *their own* domain. Platform validation succeeds → the pin is never consulted → this verifier accepts any host → the bearer token is sent to the attacker. This holds for self-signed deployments as well, because the attacker picks the cert they present.

Suggested fix — bypass hostname verification only when the pin is what established trust:

```kotlin
val defaultVerifier = OkHttpClient().hostnameVerifier // OkHostnameVerifier
baseBuilder.hostnameVerifier { hostname, session ->
    val leaf = session.peerCertificates.firstOrNull() as? X509Certificate
    val pinMatches = leaf != null && fingerprint != null &&
        Fingerprints.matches(Fingerprints.sha256(leaf), fingerprint)
    pinMatches || defaultVerifier.verify(hostname, session)
}
```

The user confirmed *that exact certificate*, so accepting it at any address is sound; everything else stays host-bound. (Also covers the `fingerprint == null` client, which currently ends up with platform-trust-only and no host binding at all.)

core-network/src/androidTest/java/io/github/xexanos/ratatoskr/network/component/FactoryTlsPinComponentTest.kt
```diff
@@ -47,6 +50,20 @@ class FactoryTlsPinComponentTest {
         assertTrue((result as ApiResult.Success).data.isEmpty())
     }
 
+    @Test
+    fun aPinnedConnectionIgnoresTheCertificateHostname() = runBlocking {
```
Xexanos, line 56: **Coverage gap:** This test (and the fixture's foreign SAN) proves hostname-independence only on the **pin path** — self-signed cert, platform chain fails, fingerprint decides — which is the safe case. The dangerous case introduced by this PR is the **platform-trusted path**: a validly chained certificate for the *wrong* host now connects, and that path is exactly the one this suite cannot reach (test CA not in the system store, as the class doc notes). If the pin-aware-verifier fix lands, a JVM-level unit test of the verifier itself (mocked `SSLSession` with a non-matching leaf + non-matching hostname → rejected) could cover it without needing a system-trusted cert.

core-network/src/testFixtures/java/io/github/xexanos/ratatoskr/network/testutil/HttpsMockServer.kt
```diff
@@ -95,14 +96,16 @@ class HttpsMockServer : ExternalResource() {
...
 
     private companion object {
-        // canonicalHostName is what MockWebServer.url() reports for the loopback address; using
-        // it as the certificate SAN keeps the served cert valid for the URL the tests call.
-        private val loopbackHost: String = InetAddress.getByName("localhost").canonicalHostName
+        // A SAN that never matches the loopback host MockWebServer serves on. The pinned client
+        // disables hostname verification (exact-cert pinning decides), so the served cert needs
+        // no matching SAN - and using a foreign one makes the whole HTTPS suite prove that
+        // independence: re-enabling the default hostname verifier would fail every test here.
+        private const val FOREIGN_SAN = "ratatoskr-e2e.invalid"
```
Xexanos, line 105: **Docs overclaim:** "a green handshake here therefore also proves the client ignores the certificate hostname" — only for the self-signed/pin path this fixture exercises. On the platform-trusted path (reverse proxy, public CA) hostname handling is untested here, and with the current factory change it is silently absent. Worth scoping the claim to the pin path so the fixture doc doesn't certify more than the suite verifies. The foreign-SAN idea itself is a good structural guard and worth keeping under a pin-aware verifier, where the pin path legitimately ignores hostnames.

Xexanos (commented):
## Review

**The problem this PR describes is real, but the fix as written opens a serious MITM hole. I'd hold off merging until the verifier is made pin-aware.**

### Overview

The PR disables hostname verification on the shared production client builder (`.hostnameVerifier { _, _ -> true }`), flips the `HttpsMockServer` fixture to serve a certificate whose SAN never matches its host, and adds an explicit regression test. Goal: let the app reach a trusted self-signed server by LAN IP / `10.0.2.2`, where the cert SAN doesn't contain the address.

### Findings

1. **Critical (security):** The PR's core premise — "exact-leaf-cert pinning guarantees the server presents the one confirmed certificate, so hostname verification is redundant" — does not match the implementation. `PinnedTrustManager` is *platform trust first, pin only as fallback* (`PinnedTrustManager.kt:50-52`, per SPEC §6's deliberate reverse-proxy trade-off). `checkServerTrusted` validates the chain to a root, **not** the hostname — host binding in OkHttp lives solely in the `HostnameVerifier` this PR disables. Result: any on-path attacker presenting a *valid publicly-trusted cert for their own domain* (free via Let's Encrypt) passes platform validation (pin never consulted) and now also passes hostname verification. TLS is established with the attacker, `BearerAuthInterceptor` sends the token → full MITM/token theft. This affects **self-signed deployments too** — the attacker chooses which cert to present. Detailed suggestion inline.

2. **Medium (test coverage):** The dangerous path — valid platform chain + non-matching host — is exactly the one the suite cannot exercise (test-minted CAs aren't in the system store, as the test doc itself notes). The suite proves hostname-independence only for the pin path, where it is safe; the SPEC §9 gap becomes security-critical with this change. Inline comment on the test.

3. **Minor (docs):** The new comments in `RatatoskrClientFactory` and `HttpsMockServer` state the "exact-cert pinning supersedes hostname verification" claim as fact; whatever fix lands, they need to describe the hybrid platform-first model accurately, or the next reader will build on the wrong mental model. The `CertificateInspector` analogy also doesn't carry: that client trusts everything and never carries real traffic, so hostname verification is meaningless there; on the production client it is the only host binding the platform path has.

### What's good

- The bug is real and well-diagnosed: default hostname verification does break LAN-IP connections to a trusted self-signed server, and it did contradict the fixture comment and E2E plans.
- The fixture strategy (serving a deliberately foreign SAN so the whole suite would fail if the verifier came back) is a nice structural guard — worth keeping once the verifier is pin-aware, since the pin path legitimately ignores hostnames.
- `ratatoskr-e2e.invalid` uses an RFC-2606-reserved TLD; good choice.
- Doc comments updated in lockstep with behaviour, per repo convention.

### Suggested direction

Make hostname bypass conditional on the pin actually having decided: a custom `HostnameVerifier` that accepts when the presented leaf matches the confirmed fingerprint, and otherwise falls back to OkHttp's default verifier. That keeps the reverse-proxy/public-CA deployment host-bound while giving the confirmed self-signed cert address-independence — the behaviour the PR description already claims. Alternative (server-side): include the server's IP/hostname in the generated cert's SAN (ratatoskr-server#42 mints it anyway).

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientFactory.kt
Xexanos: Confirmed - the premise was inverted, thanks for catching it before merge. Implemented as suggested in 9f973e2: `PinnedHostnameVerifier` skips the hostname check only when the served leaf matches the confirmed fingerprint; everything else (platform-trusted path, and the `fingerprint == null` client) delegates to OkHttp's standard verifier. It also fails closed when `peerCertificates` throws `SSLPeerUnverifiedException`. The instrumented foreign-SAN suite stays green through the real Conscrypt stack, so the verifier does get the peer certificates it needs on device.

core-network/src/androidTest/java/io/github/xexanos/ratatoskr/network/component/FactoryTlsPinComponentTest.kt
Xexanos: Added `PinnedHostnameVerifierTest` (JVM, 9f973e2) as suggested: a fake `SSLSession` serving real `HeldCertificate`s. It covers the attack case (non-matching leaf + non-matching hostname -> rejected), the no-pin client staying host-bound, delegation to the default verifier on a matching SAN, the LAN-IP pin case, and an unverified peer failing closed.

core-network/src/testFixtures/java/io/github/xexanos/ratatoskr/network/testutil/HttpsMockServer.kt
Xexanos: Scoped in 9f973e2: the fixture doc (and `FactoryTlsPinComponentTest`) now claim hostname-independence of the pin path only, and point to `PinnedHostnameVerifierTest` for the platform path's host binding. The foreign SAN itself stays as the structural guard.

PR #36, opened 2026-07-15
feat(connect): tag the server-URL field for black-box E2E driving

Xexanos:
The first E2E run surfaced that the connect screen's **server-URL field** can't be driven
reliably black-box: it's prefilled (`https://`) with only a floating label, so Maestro tapping
the "Server URL" label lands the cursor mid-prefill → erase+type produced `https://localhost:8080s://`
→ "Invalid URL port" (screenshot in the e2e run).

Give the field a stable `testTag` (`connect_server_url`) so a black-box driver targets the field
node directly. This is exactly the case `UiTestTags` already documents — tag the elements that have
no reliable text/semantics handle (the URL field joins the search box, list rows, and the slider).

- `UiTestTags.CONNECT_SERVER_URL = "connect_server_url"`
- applied to the `OutlinedTextField` in `ConnectScreen.kt`

No behavior change; `testTagsAsResourceId` (app#15) already exposes it to UiAutomator/Maestro.
Once merged + a new `testing-<sha>` APK is published, the E2E flow switches to `id: connect_server_url`.

### DIFF
--- a/app/src/main/java/io/github/xexanos/ratatoskr/ui/UiTestTags.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/ui/UiTestTags.kt
@@ -13,8 +13,6 @@ package io.github.xexanos.ratatoskr.ui
  * existing text/contentDescription instead - see the integration suite.
  */
 object UiTestTags {
-    /** Connect screen's server-URL field: prefilled ("https://") with only a floating label,
-     *  so it has no reliable text handle for black-box driving (the central E2E harness). */
     const val CONNECT_SERVER_URL = "connect_server_url"
     const val LIBRARY_SEARCH = "library_search"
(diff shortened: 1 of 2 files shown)


app/src/main/java/io/github/xexanos/ratatoskr/ui/UiTestTags.kt
```diff
@@ -13,6 +13,9 @@ package io.github.xexanos.ratatoskr.ui
  * existing text/contentDescription instead - see the integration suite.
  */
 object UiTestTags {
+    /** Connect screen's server-URL field: prefilled ("https://") with only a floating label,
+     *  so it has no reliable text handle for black-box driving (the central E2E harness). */
```
Xexanos:
```suggestion
```

PR #37, opened 2026-07-15
ci: arm the E2E trigger, add the app release-promote gate, enforce version bumps

Xexanos:
Arms the app side of the cross-component pipeline and automates release tagging via the E2E gate. Structure deliberately modeled on the server repo (`container.yml` → E2E → `promote.yml`).

## What's included

**1. E2E trigger armed** (`release-validation.yml`)
The placeholder echo is gone: on a green on-device validation the workflow dispatches `app-validated` to `Xexanos/ratatoskr-e2e` (carrying the `testing-<sha>` tag + commit). Needs the `E2E_DISPATCH_TOKEN` secret; the step fails loudly if it is missing.

**2. App release gate** (new: `promote.yml`)
Triggered by the `app-e2e-passed` dispatch the E2E side fires back on a green run. Cuts an annotated `v<versionName>` tag at the validated commit plus a GitHub release carrying the E2E-validated unsigned APK (copied from the `testing-<sha>` pre-release, **no rebuild**) — but only if that `versionName` is not already tagged. `versionName` stays the human-maintained source of truth (SPEC §8); only the "wait for green, then tag" half is automated. F-Droid rebuilds from the tag and signs itself.

**3. Version-bump enforcement** (new: `pr-release-guards.yml` + `commitlint.config.mjs`)
- `commitlint` — Conventional Commits on every PR commit (mirrored from the server).
- `version-bump` — for release-worthy commits (feat/fix/perf/breaking) the PR **must** raise `versionName`/`versionCode` by at least the implied semver level, keep the `MAJOR*10000+MINOR*100+PATCH` formula, and ship the `en-US` changelog; otherwise it goes red. Catches a forgotten bump at merge time instead of letting `promote.yml` silently no-op later.
- `promote.yml` now warns loudly (instead of a silent notice) when it lands on an already-tagged `versionName` that has release-worthy commits since (the race / bypassed-guard case).

**4. SPEC §8** now describes the automated, E2E-gated tagging, the bump enforcement, and *why* the version must be set before the build on Android (compiled into the manifest + signed — no "retag" as with OCI images).

**5. Version baseline set to 1.0.0** (`chore(release)`)
`versionName` `0.1.0` → `1.0.0`, `versionCode` `100` → `10000`; the fastlane changelog is renamed `100.txt` → `10000.txt` (en-US + de-DE) and reworded to "initial release". Safe to re-baseline: `0.1.0` was never released (no `v*` tag exists, only throwaway `testing-*` pre-releases).

## Steps required outside this PR

- [ ] **E2E repo**: apply the counterpart patch (accepts `app-validated` + dispatches back). Patch is ready; see the session handoff.
- [ ] Secret `E2E_DISPATCH_TOKEN` (this repo) — PAT with access to `ratatoskr-e2e`.
- [ ] Secret `APP_PROMOTE_TOKEN` (E2E repo) — PAT allowed to dispatch this repo.
- [ ] Branch protection on `main`: make `Conventional Commits` + `Version bump required for release-worthy changes` **required checks**; enable "require branches to be up to date before merging". **Disable squash merges** (use merge-commit or rebase) so Conventional Commits reach `main` - see SPEC §8.

## Verification

YAML of all workflows validated; the guard bash (classification, formula, semver comparison) tested locally. The real dispatch loop is only testable live once the branch is pushed and the secrets exist.

## Note on the server template

The template was `promote.yml` from server **PR #55** (`promote-auto-semver`, still open), not server `main`. Comments stating "the server derives semver from Conventional Commits" assume #55 lands.

### DIFF
--- a/.github/workflows/pr-release-guards.yml
+++ b/.github/workflows/pr-release-guards.yml
@@ -1,0 +1,124 @@
+# Per-PR release guards. Two independent checks that keep the continuous-versioning release model
+# (SPEC section 8) honest, so a release-worthy change can never merge without the version bump the
+# automated tag gate (promote.yml) needs - "forgetting to bump" is caught here, loudly, instead of
+# silently producing no release later (promote.yml no-ops when the versionName is already tagged).
+#
+#   - commitlint    : every PR commit must be a Conventional Commit (mirrors the server repo). The
+#                     TYPES are load-bearing - the version-bump guard classifies them to decide
+#                     whether a bump is required. Rules live in commitlint.config.mjs at the root.
+#   - version-bump  : if the PR carries release-worthy commits (feat/fix/perf/breaking) it MUST
+#                     bump versionName/versionCode in app/build.gradle.kts by at least the semver
+#                     level those commits imply, keep versionCode consistent with the SPEC formula,
+#                     and ship the matching fastlane changelog. Otherwise no bump is required.
+#
+# Deliberately its own workflow (no paths-ignore): both jobs must ALWAYS report so they are safe to
+# make required status checks. A docs-only PR still runs them - commitlint always applies, and the
+# version-bump guard simply finds no release-worthy commits and passes.
+#
+# Concurrent-PR caveat: the guard compares against the PR's base. Two PRs branched off the same
+# commit can both bump to the same version and both pass; whichever merges second then re-points no
+# new tag. Enable branch protection's "require branches to be up to date before merging" so the
+# second PR must rebase onto the new main and its guard re-computes the next free version.
+
+name: PR release guards
+
+on:
+  pull_request:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: pr-release-guards-${{ github.head_ref }}
+  cancel-in-progress: true
+
+jobs:
+  commitlint:
+    name: Conventional Commits
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          # The PR's commits must be present locally to lint the base..head range.
+          fetch-depth: 0
+      - uses: actions/setup-node@v6.4.0
+        with:
+          node-version: 22
+      - name: Lint the PR's commit messages
+        # npx with pinned versions (not a devDependency): this is a Kotlin/Gradle repo with no
+        # package.json, so the workspace stays JS-free. Pins match the server repo's commitlint job.
+        run: |
+          npx --yes -p @commitlint/cli@21.2.1 -p @commitlint/config-conventional@21.2.0 \
+            commitlint --from ${{ github.event.pull_request.base.sha }} \
+                       --to ${{ github.event.pull_request.head.sha }} --verbose
+
+  version-bump:
+    name: Version bump required for release-worthy changes
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v7
+        with:
+          # Need the base..head commit range and both revisions' app/build.gradle.kts.
+          fetch-depth: 0
+      - name: Require a version bump when the PR is release-worthy
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+
+          # --- classify the PR's commits (identical rules to the server's promote.yml) ----------
+          subjects="$(git log --format='%s' "${BASE_SHA}..${HEAD_SHA}")"
+          bodies="$(git log --format='%b' "${BASE_SHA}..${HEAD_SHA}")"
+          bump="none"
+          if grep -qE '^[a-z]+(\([^)]*\))?!:' <<< "$subjects" \
+             || grep -qE '^BREAKING[ -]CHANGE:' <<< "$bodies"; then
+            bump="major"
+          elif grep -qE '^feat(\([^)]*\))?:' <<< "$subjects"; then
+            bump="minor"
+          elif grep -qE '^(fix|perf)(\([^)]*\))?:' <<< "$subjects"; then
+            bump="patch"
+          fi
+
+          if [ "$bump" = "none" ]; then
+            echo "No release-worthy commits (feat/fix/perf/breaking) in this PR - a version bump is not required."
+            exit 0
+          fi
+
+          # --- read versionName/versionCode at base and head ------------------------------------
+          read_vn() { git show "$1:app/build.gradle.kts" | grep -E '^\s*versionName\s*=' | head -n1 | grep -oP 'versionName\s*=\s*"\K[^"]+'; }
+          read_vc() { git show "$1:app/build.gradle.kts" | grep -E '^\s*versionCode\s*='  | head -n1 | grep -oP 'versionCode\s*=\s*\K[0-9]+'; }
+          base_vn="$(read_vn "$BASE_SHA")"; base_vc="$(read_vc "$BASE_SHA")"
+          head_vn="$(read_vn "$HEAD_SHA")"; head_vc="$(read_vc "$HEAD_SHA")"
+
+          fail() { echo "::error::$*"; exit 1; }
+
+          case "$head_vn" in
+            [0-9]*.[0-9]*.[0-9]*) : ;;
+            *) fail "versionName '${head_vn}' is not MAJOR.MINOR.PATCH (SPEC section 8)." ;;
+          esac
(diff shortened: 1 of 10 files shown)


.github/workflows/promote.yml
```diff
@@ -0,0 +1,174 @@
...
+
+          # Idempotent gate: cut the tag only if this version has not been released yet.
+          if git rev-parse -q --verify "refs/tags/v${version}" >/dev/null; then
+            # Distinguish the benign case (a non-release commit - docs/chore/ci - promoted past an
+            # already-released version) from the one that means a bump was MISSED: release-worthy
+            # commits between v${version} and this commit. The PR guard (pr-release-guards.yml)
+            # should stop the latter at merge time, so reaching it here signals the guard was
+            # bypassed or two PRs collided on the same version - warn loudly rather than no-op
+            # silently (the whole point of the guard is that forgetting a bump is never quiet).
+            subjects="$(git log --format='%s' "v${version}..${commit}")"
```
Xexanos, line 130: **Merge strategy vs. main-history classification (medium).** The version-bump guard classifies the *PR commits* (`base..head`) — correct. But this block classifies `git log v${version}..${commit}` over the **main history**. With **squash merges**, the main commit's subject defaults to the **PR title**, and PR titles are not linted (commitlint only checks the branch commits). A `feat` PR titled `ci: …` then lands as a non-release-worthy commit on main, so a genuinely missed bump could be reported as the benign `::notice::` instead of the `::warning::`. Suggest pinning the merge strategy (merge-commit/rebase) or additionally linting the squash title, and recording the chosen strategy in SPEC §8 — the whole flow implicitly depends on it.

.github/workflows/pr-release-guards.yml
```diff
@@ -0,0 +1,124 @@
...
+            bump="patch"
+          fi
+
+          if [ "$bump" = "none" ]; then
+            echo "No release-worthy commits (feat/fix/perf/breaking) in this PR - a version bump is not required."
+            exit 0
+          fi
+
+          # --- read versionName/versionCode at base and head ------------------------------------
+          read_vn() { git show "$1:app/build.gradle.kts" | grep -E '^\s*versionName\s*=' | head -n1 | grep -oP 'versionName\s*=\s*"\K[^"]+'; }
```
Xexanos: **Opaque failure under `pipefail` (low).** If `versionName`/`versionCode` (or the file) is missing at the revision, the trailing `grep` exits 1; under `set -euo pipefail` the script aborts inside the command substitution on line 91/92 *before* reaching the nice `fail`/format checks below — the user sees a bare non-zero exit instead of "Could not read versionName …". Consider checking for empty right after reading and calling `fail` with a clear message.

.github/workflows/pr-release-guards.yml
```diff
@@ -0,0 +1,124 @@
...
+          fetch-depth: 0
+      - name: Require a version bump when the PR is release-worthy
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+
+          # --- classify the PR's commits (identical rules to the server's promote.yml) ----------
+          subjects="$(git log --format='%s' "${BASE_SHA}..${HEAD_SHA}")"
```
Xexanos, line 74: Minor/info: `revert` is release-neutral in this classifier, so a revert of a shipped `feat`/`fix` would pass without a bump. Rarely relevant and presumably intentional — just flagging for awareness.

commitlint.config.mjs
```diff
@@ -0,0 +1,41 @@
...
+        'deps', // dependency bumps (dependabot's configured prefix); use fix(deps): when a runtime bump should ship a release
+        'review', // addressing PR review feedback
+        'spec', // docs/SPEC.md changes
+        'comment', // comment-only corrections
+      ],
+    ],
+    // Style rules relaxed for dependabot, which capitalizes subjects ("deps: Bump ...")
+    // and writes long changelog URLs into bodies. The gate exists so the TYPE parses -
+    // subject cosmetics are not what the release pipeline reads.
+    'subject-case': [0],
```
Xexanos, line 39: **Comment framing (low).** These three rules (`subject-case`, `body-max-line-length`, `header-max-length`) are disabled **globally**, but the comment reads as if the relaxation is Dependabot-only. It's a deliberate trade-off (only the `type` is load-bearing), but the comment should make clear the checks are off for *all* commits — otherwise it will later surprise someone that 100+ char headers pass in general.

.github/workflows/release-validation.yml
```diff
@@ -171,21 +173,32 @@ jobs:
...
+          RELEASE_TAG: ${{ steps.tag.outputs.name }}
+          SHA: ${{ github.sha }}
+        run: |
+          if [ -z "$E2E_DISPATCH_TOKEN" ]; then
+            echo "::error title=E2E dispatch failed::E2E_DISPATCH_TOKEN is not set - cannot trigger the E2E suite for ${RELEASE_TAG}. Add the secret (a PAT with access to ratatoskr-e2e)." >&2
+            exit 1
+          fi
+          echo "Dispatching app-validated E2E for ${RELEASE_TAG} (${SHA})"
+          GH_TOKEN="$E2E_DISPATCH_TOKEN" gh api \
+            repos/${{ github.repository_owner }}/ratatoskr-e2e/dispatches \
```
Xexanos: Cosmetic: `github.repository_owner` is trusted, so no injection risk, but interpolating `${{ … }}` directly into the `run:` script is inconsistent with the rest of this step, which cleanly passes values via `env:` (`RELEASE_TAG`, `SHA`). Consider passing the owner through `env:` too for consistency.

Xexanos (commented):
Overall a very clean, exceptionally well-documented CI PR — the comments consistently explain the *why*, the Bash guards are defensive (`set -euo pipefail`, fetch-bytes-before-tag, `cancel-in-progress: false` on promotion), and the mirroring of the server pipeline is consistent. No blockers. The inline notes below are mostly nits / operational caveats. Separately verified as fine: `target_commitish: github.sha` makes promote's manual path resolve a real SHA; download-before-tag ordering; the `app-promote` concurrency gate; tag push via `GITHUB_TOKEN` not re-triggering workflows; no secret logging; `checkout@v7` consistent repo-wide.

.github/workflows/promote.yml
Xexanos: Good catch. The guard's own `base..head` classification is strategy-independent, but this main-history diagnostic isn't. Documented the assumption in the code comment and recorded the no-squash requirement in SPEC §8; also added "disable squash merges" to the out-of-band checklist in the PR description. Fixed in 758c5e4.

.github/workflows/pr-release-guards.yml
Xexanos: Fixed in 758c5e4: the reads now use `|| true` and an empty-check that calls `fail()` with a clear message right after, and `fail()` is moved above the reads so it's in scope.

.github/workflows/pr-release-guards.yml
Xexanos: Intentional — kept release-neutral to match the server's classifier, and now called out in a code comment so it isn't mistaken for an oversight (758c5e4).

commitlint.config.mjs
Xexanos: Reworded in 758c5e4 to make explicit that all three style rules are disabled globally (only the commit type is load-bearing), with Dependabot as the motivating example rather than the scope.

.github/workflows/release-validation.yml
Xexanos: Done in 758c5e4 — the owner is now passed via `env:` (`OWNER`) like the other values.

Xexanos:
Pushed a small commit here (`ci: dispatch app-candidate…`) as part of unifying the four cross-repo dispatch event names to a consistent `<component>-<phase>` scheme:

- `app-validated` → **`app-candidate`** (this PR's E2E-trigger dispatch)
- callback `app-e2e-passed` unchanged

Counterparts: **ratatoskr-e2e#3** (now listens for `app-candidate` / `server-candidate`, sends `server-e2e-passed` / `app-e2e-passed`) and **ratatoskr-server#63** (`server-candidate` / `server-e2e-passed`).

⚠️ `repository_dispatch` `event_type` is unvalidated, so this PR must merge together with ratatoskr-e2e#3 or the app→E2E trigger silently no-ops. Shout if you'd rather own this line yourself and I'll drop the commit.

PR #38, opened 2026-07-16
fix(nowplaying): clear a stale error when the session ends

Xexanos:
## Problem

In `NowPlayingViewModel.refresh()`, a `NoActiveSession` (404) poll result dropped the session card but **kept the last error banner**:

```kotlin
is RatatoskrError.NoActiveSession ->
    _uiState.value = _uiState.value.copy(loading = false, session = null)   // error left set
```

So when a speaker drops out mid-session — the poll returns 502 and the screen shows "Sonos is unavailable" — and the server then **relinquishes the session** (next poll 404), the app cleared the card but left the stale banner, showing "Sonos is unavailable" on an otherwise-empty Now-playing screen instead of "Nothing is playing right now."

`applySession()` already clears a stale error on success for exactly this "no sticky banner" reason; the 404 branch was the one spot that didn't.

## Fix

Clear the error alongside the session in the `NoActiveSession` branch. One line, plus a regression test (`refresh clears a stale error once the session has ended`): a 502 sets the banner, a subsequent 404 must leave `session == null` **and** `error == null`.

## How it was caught

The E2E-09 recovery assertion in [Xexanos/ratatoskr-e2e#2](https://github.com/Xexanos/ratatoskr-e2e/pull/2) (speaker disappears mid-session → comes back empty → server relinquishes): the server returned 404 exactly as designed, but the app UI stuck on the old "Sonos is unavailable" banner, so the flow's "Nothing is playing right now." assertion failed. Verified from the server log (steady 404s) and the failure screenshot (empty Now-playing with the stale banner).

Once this lands and ships in a `testing-*` APK, ratatoskr-e2e#2's E2E-09 goes green end-to-end.

### DIFF
--- a/app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
@@ -111,7 +111,11 @@ class NowPlayingViewModel(
                 if (epoch != commandEpoch || _uiState.value.stopped) return
                 when (result.error) {
                     is RatatoskrError.NoActiveSession ->
-                        _uiState.value = _uiState.value.copy(loading = false, session = null)
+                        // Session ended (server relinquished / stopped): drop the card AND clear any
+                        // error, so a banner from a just-prior failure (e.g. a 502 while the speaker
+                        // was dropping out) doesn't stick on the now-empty screen. Mirrors the
+                        // stale-error clear in applySession().
+                        _uiState.value = _uiState.value.copy(loading = false, session = null, error = null)
                     else ->
                         _uiState.value = _uiState.value.copy(loading = false, error = result.error.toMessage())
                 }
--- a/app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
@@ -110,12 +110,18 @@ class NowPlayingViewModel(
                 // A control action issued while this poll was in flight takes precedence.
                 if (epoch != commandEpoch || _uiState.value.stopped) return
                 when (result.error) {
-                    is RatatoskrError.NoActiveSession ->
-                        // Session ended (server relinquished / stopped): drop the card AND clear any
-                        // error, so a banner from a just-prior failure (e.g. a 502 while the speaker
-                        // was dropping out) doesn't stick on the now-empty screen. Mirrors the
-                        // stale-error clear in applySession().
+                    is RatatoskrError.NoActiveSession -> {
+                        // The session ended (server relinquished it, or it was stopped elsewhere).
+                        // The server owns token rotation only WHILE a session is active (SPEC §5),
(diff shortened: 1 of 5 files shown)


Xexanos (commented):
Clean, minimal fix — correct diagnosis, one-line change at exactly the right spot, and the regression test reproduces the real E2E-09 sequence (502 sets the banner, 404 must clear it) rather than testing the branch in isolation. The comment mirrors the existing `applySession()` rationale, matching the codebase's comment style. Verified against the surrounding code: the epoch/`stopped` guards still apply before this branch, so a control action in flight during the 404 poll drops the result as before — clearing the error here cannot race a fresh control error. One adjacent (pre-existing) gap flagged inline; one release-flow interaction note below.

**Interaction with PR #37 (non-blocking):** this is a `fix:` commit. For the E2E purpose stated in the description a `testing-*` APK is enough (per-commit, no version needed). But under the release model #37 arms, a `fix:` landing without a `versionName` bump ships in no release — and once #37's guard is a required check, this PR would be red without a bump + changelog. Worth deciding merge order / whether this fix should carry `0.x.y+1` now.

app/src/main/java/io/github/xexanos/ratatoskr/ui/nowplaying/NowPlayingScreen.kt
```diff
@@ -111,7 +111,11 @@ class NowPlayingViewModel(
                 if (epoch != commandEpoch || _uiState.value.stopped) return
                 when (result.error) {
                     is RatatoskrError.NoActiveSession ->
-                        _uiState.value = _uiState.value.copy(loading = false, session = null)
+                        // Session ended (server relinquished / stopped): drop the card AND clear any
+                        // error, so a banner from a just-prior failure (e.g. a 502 while the speaker
+                        // was dropping out) doesn't stick on the now-empty screen. Mirrors the
+                        // stale-error clear in applySession().
+                        _uiState.value = _uiState.value.copy(loading = false, session = null, error = null)
```
Xexanos, line 124: **Adjacent gap, pre-existing (suggestion):** `applySession()` updates `connectionManager.setSessionActive(...)` on every success, and `stop()` sets it `false` — but this `NoActiveSession` branch never does. In exactly the scenario this PR fixes (server relinquishes the session), the app keeps `sessionActive = true` for as long as the user sits on the now-empty screen polling 404s, which suppresses the client's own token refresh (SPEC §5: the server owns rotation only *while a session is active* — and there is none anymore). Since this PR touches this exact branch, consider adding `connectionManager.setSessionActive(false)` here (plus an assertion in the new test). Fine as a follow-up too if you want to keep this PR one-line.

Xexanos: Good catch — done in 39c0e97 (not a follow-up). Set `setSessionActive(false)` in this branch so a relinquished session stops suppressing the client's own refresh; that's actually the root of the token-lapse we saw in E2E-09's recovery timing. Added `ConnectionManager.isSessionActive()` and folded the assertion into the regression test as the full sequence: playing → 502 banner (session + flag retained) → 404 relinquish (card gone, banner cleared, flag released).

PR #45, opened 2026-07-16
ci: publish a developer-signed sideload APK on release

Xexanos:
## Why

Downloading the release APK from GitHub fails to install ("the package appears to be invalid"): the attached APK is unsigned, and Android refuses unsigned APKs. The old promote.yml comment calling it "a convenience for sideloaders" was factually wrong. Sideloading is now offered as an explicit alternative to F-Droid.

## What

- **promote.yml**: new "Sign the sideload APK" step — zipaligns and signs the E2E-validated unsigned APK (the exact bytes that passed, never a rebuild) with a developer release key from the new `RELEASE_KEYSTORE` (base64 PKCS12) / `RELEASE_KEYSTORE_PASSWORD` secrets, verifies the result, and attaches `ratatoskr-v<versionName>.apk` alongside the unsigned provenance APK. Runs before the tag push, so a missing secret fails the run before any tag exists. Release notes now carry install guidance including the channel-exclusivity warning.
- **SPEC section 8**: the "F-Droid signs its own builds" decision becomes "two install channels, two keys", recording the trade-offs: signature continuity makes F-Droid and sideload mutually exclusive per install, and there is now a keystore to manage and back up. F-Droid's path is unchanged (it consumes only the tag).

## Notes

- The two repo secrets must be set before the next promotion runs, otherwise the sign step fails (intentionally, before tagging).
- The existing v1.0.0 release has already been fixed out-of-band: its validated APK was signed with the same key and uploaded as `ratatoskr-v1.0.0.apk`, and the release notes got the same install section.
- Certificate SHA-256: `1B:4D:89:B7:B6:CF:20:70:22:A2:38:25:E5:65:1B:F5:6B:4F:CF:CA:EC:BB:35:10:97:72:30:42:66:F0:11:B1`
- Not testable in CI until the next version bump promotes (the idempotent gate no-ops on already-released versions); the signing commands themselves were exercised locally against the v1.0.0 bytes.

### DIFF
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -215,7 +215,7 @@ jobs:
             --notes "## Install
 
           - **F-Droid** (primary): builds and signs this release itself from the \`v${VERSION}\` tag.
-          - **Sideload**: download \`ratatoskr-v${VERSION}.apk\` (developer-signed; certificate SHA-256 \`D9:91:F4:BC:63:BB:23:F5:87:F0:58:57:10:65:AC:B2:00:3E:61:BC:A7:DD:12:95:BB:C2:21:7D:62:87:AD:DF\`).
+          - **Sideload**: download \`ratatoskr-v${VERSION}.apk\` (developer-signed; certificate SHA-256 \`1B:4D:89:B7:B6:CF:20:70:22:A2:38:25:E5:65:1B:F5:6B:4F:CF:CA:EC:BB:35:10:97:72:30:42:66:F0:11:B1\`).
 
           The two channels are signed with different keys: a device cannot switch between an F-Droid install and a sideloaded install without uninstalling first. \`ratatoskr-release-unsigned.apk\` is the unsigned build the E2E suite validated (provenance artifact, not installable)." \
             --generate-notes \
--- a/.github/workflows/promote.yml
+++ b/.github/workflows/promote.yml
@@ -5,8 +5,14 @@
 # what passed. This mirrors the server's promote.yml (which re-tags the tested image by digest);
 # an APK cannot be "re-tagged" the way a container image can (versionName/versionCode are compiled
 # into the APK bytes and covered by the signature), so the tag is what carries the version, and
-# F-Droid rebuilds from that tag and signs with its own key (SPEC section 8). The attached APK is
-# a convenience for sideloaders, not what F-Droid consumes.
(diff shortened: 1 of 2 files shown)


.github/workflows/promote.yml
```diff
@@ -159,6 +165,34 @@ jobs:
...
+          printf '%s' "$KEYSTORE_B64" | base64 -d > "$keystore"
+          bt="$ANDROID_HOME/build-tools/$(ls "$ANDROID_HOME/build-tools" | sort -V | tail -n1)"
+          # zipalign before signing (v2+ signatures cover the whole file). AGP's unsigned APK
+          # is already aligned; -f makes this idempotent rather than a failure.
+          "$bt/zipalign" -f 4 publish/ratatoskr-release-unsigned.apk "$RUNNER_TEMP/aligned.apk"
+          # --ks-pass env: keeps the password out of argv and logs (never log secrets).
+          "$bt/apksigner" sign \
+            --ks "$keystore" --ks-key-alias ratatoskr-release --ks-pass env:KEYSTORE_PASS \
+            --out "publish/ratatoskr-v${VERSION}.apk" "$RUNNER_TEMP/aligned.apk"
+          "$bt/apksigner" verify --print-certs "publish/ratatoskr-v${VERSION}.apk"
```
Xexanos: `verify --print-certs` only *prints* the certificate — it never asserts it. If the wrong keystore ends up in the `RELEASE_KEYSTORE` secret (or a rotated one goes stale), this step signs happily with a foreign key while the release notes still claim the pinned fingerprint. For existing sideload installs that release would be an uninstallable update — exactly the failure the channel-exclusivity warning is about. Since the expected fingerprint is already hardcoded in the notes below, asserting it here is cheap:

```suggestion
          "$bt/apksigner" verify --print-certs "publish/ratatoskr-v${VERSION}.apk" \
            | grep -Fq 'certificate SHA-256 digest: 1b4d89b7b6cf207022a23825e5651bf56b4fcfcaecbb35109772304266f011b1' \
            || { echo "::error::Signing certificate does not match the pinned fingerprint - wrong or rotated RELEASE_KEYSTORE secret."; exit 1; }
```

(apksigner prints the digest lowercase without colons.) On key rotation this constant, the fingerprint in the release notes, and SPEC section 8 all need to change together — the assertion makes forgetting one of them fail loudly instead of shipping a mislabeled release.

PR #47, opened 2026-07-16
feat: add a German translation and per-app language config

Xexanos:
## Why

The app ships only default (English) strings, so a German-locale device stays English — the localization plumbing (all screens use `stringResource`) was fine, the translations simply did not exist.

## What

- **`values-de/strings.xml`**: German translation of all 58 translatable strings. Terminology follows the official Audiobookshelf app's German translation (informal du-form); `library_percent` uses a non-breaking space before `%` per German typography.
- **`values/strings.xml`**: `app_name` and the `-` unknown-state placeholder are marked `translatable="false"` (and therefore absent from `values-de`).
- **`locales_config.xml` + manifest `android:localeConfig`**: declares en + de so Android 13+ lists the app in the per-app language settings; ignored on older releases.
- **Version bump 1.0.1 → 1.1.0** (versionCode 10100) with fastlane changelogs (en-US, de-DE): user-visible feature, release-worthy per the PR guard.

## Notes

- Merging this releases v1.1.0 through the promote pipeline — the first release to exercise the new sideload-signing step from #45 end to end.
- Verified locally: `:app:assembleDebug` builds and the merged `values-de` resources contain the app's German strings.

### DIFF
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -18,7 +18,7 @@
 
     <!-- Sign in -->
     <string name="signin_title">Bei Audiobookshelf anmelden</string>
-    <string name="signin_subtitle">Verwende deinen vorhandenen Benutzernamen und dein Passwort.</string>
+    <string name="signin_subtitle">Verwende deinen bestehenden Benutzernamen und dein Passwort.</string>
     <string name="signin_username_label">Benutzername</string>
     <string name="signin_password_label">Passwort</string>
     <string name="signin_action">Anmelden</string>
--- a/app/src/main/res/values-de/strings.xml
+++ b/app/src/main/res/values-de/strings.xml
@@ -1,0 +1,71 @@
+<resources>
+    <!-- Connect -->
+    <string name="connect_welcome_title">Willkommen bei Ratatoskr</string>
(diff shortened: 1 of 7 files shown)


app/src/main/res/values-de/strings.xml
```diff
@@ -0,0 +1,71 @@
...
+    <string name="connect_cert_issuer_label">Aussteller</string>
+    <string name="connect_cert_valid_until_label">Gültig bis</string>
+    <string name="connect_cert_fingerprint_label">SHA-256-Fingerabdruck</string>
+    <string name="connect_cert_compare_hint">Vergleiche diesen Fingerabdruck mit dem, den dein Server anzeigt, bevor du ihm vertraust.</string>
+    <string name="connect_action_trust">Vertrauen und fortfahren</string>
+    <string name="connect_action_cancel">Abbrechen</string>
+
+    <!-- Sign in -->
+    <string name="signin_title">Bei Audiobookshelf anmelden</string>
+    <string name="signin_subtitle">Verwende deinen vorhandenen Benutzernamen und dein Passwort.</string>
```
Xexanos:
```suggestion
    <string name="signin_subtitle">Verwende deinen bestehenden Benutzernamen und dein Passwort.</string>
```

Xexanos: Done in f397e8a.

PR #48, opened 2026-07-17
ci: retry the flaky emulator-package download before android-emulator-runner

Xexanos:
## Summary
- sdkmanager's Android Emulator package download intermittently arrives as a corrupted zip ("Error on ZipFile unknown archive"), which `reactivecircus/android-emulator-runner` treats as fatal — PR #43 lost a full run to exactly this.
- Adds a pre-install step (`sdkmanager --install emulator --channel=0`, retried up to 3x with a 10s backoff) before every call to the action, in both `instrumented-tests.yml` (6 jobs) and `release-validation.yml` (1 job). If our install succeeds, the action's own install of the same package is a no-op; if all 3 attempts fail, the action still gets one more try and fails visibly as before.
- Included `release-validation.yml` too, since a flake there reports as a main-branch failure rather than a PR failure.

## Test plan
- [ ] Open this PR and confirm the `Instrumented tests` checks pass (exercises the new pre-install step on every job that hits it).
- [ ] Spot-check one job's log for the new "Pre-install the Android Emulator SDK package" step to confirm it runs and, on success, that the action's own "Install Android SDK" step reports the package already present.

### DIFF
--- a/.github/actions/preinstall-emulator/action.yml
+++ b/.github/actions/preinstall-emulator/action.yml
@@ -20,13 +20,24 @@ runs:
   steps:
     - shell: bash
       run: |
-        # Not on PATH yet at this point in the job - the emulator-runner action's own
-        # JS wrapper is what adds it. Resolve the binary directly under $ANDROID_HOME.
-        sdkmanager="$(find "$ANDROID_HOME/cmdline-tools" -maxdepth 2 -name sdkmanager -type f | head -n1)"
-        if [ -z "$sdkmanager" ]; then
-          echo "::warning::sdkmanager not found under \$ANDROID_HOME/cmdline-tools; skipping the pre-install."
+        # sdkmanager is NOT on PATH here - the emulator-runner action's own JS wrapper is
+        # what adds it later, so resolve it explicitly. It lives at
+        # cmdline-tools/latest/bin/sdkmanager - the exact path android-emulator-runner
+        # expects to be populated. Anchor to that; fall back to a versioned cmdline-tools
+        # dir (find -maxdepth 3 reaches <version>/bin/sdkmanager, which is three levels
+        # below cmdline-tools) only if `latest` is absent.
+        sdkmanager="$ANDROID_HOME/cmdline-tools/latest/bin/sdkmanager"
+        if [ ! -x "$sdkmanager" ]; then
+          sdkmanager="$(find "$ANDROID_HOME/cmdline-tools" -maxdepth 3 -name sdkmanager -type f 2>/dev/null | head -n1)"
+        fi
+        if [ -z "$sdkmanager" ] || [ ! -x "$sdkmanager" ]; then
+          echo "::warning::sdkmanager not found under \$ANDROID_HOME/cmdline-tools; skipping the pre-install (the emulator-runner action will still try its own install)."
           exit 0
         fi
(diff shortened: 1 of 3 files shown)


.github/actions/preinstall-emulator/action.yml
```diff
@@ -0,0 +1,40 @@
...
+name: Pre-install the Android Emulator SDK package
+description: Retry sdkmanager's flaky emulator-package download before android-emulator-runner hits it.
+runs:
+  using: composite
+  steps:
+    - shell: bash
+      run: |
+        # Not on PATH yet at this point in the job - the emulator-runner action's own
+        # JS wrapper is what adds it. Resolve the binary directly under $ANDROID_HOME.
+        sdkmanager="$(find "$ANDROID_HOME/cmdline-tools" -maxdepth 2 -name sdkmanager -type f | head -n1)"
```
Xexanos: **Critical — defeats this PR's stated goal.** `find "$ANDROID_HOME/cmdline-tools" -maxdepth 2 -name sdkmanager -type f` will never match. The real binary lives at `cmdline-tools/latest/bin/sdkmanager`, which is 3 levels below `cmdline-tools` (`latest`=1, `bin`=2, `sdkmanager`=3), not 2 — confirmed against `android-emulator-runner`'s own `sdk-installer.ts` (which builds exactly that path) and with a local `find -maxdepth` reproduction. Because the "not found" branch (lines 26-29) always fires and `exit 0`s, this step silently no-ops on every job run: it prints a warning and reports success, but the retry loop (lines 31-39) never actually executes. The step looks green but never does the thing it was built to do. Needs `-maxdepth 3` (or drop the depth limit).

Minor, same line, latent once the above is fixed: even with the right maxdepth, `find ... | head -n1` has no deterministic ordering guarantee if more than one `sdkmanager` binary exists under `cmdline-tools` (e.g. a stale/alternate version dir alongside `latest`). Consider anchoring to `cmdline-tools/latest/bin/sdkmanager` explicitly, since that's the exact path `android-emulator-runner` itself expects to already be populated.

Xexanos (commented):
Automated review findings (see inline comments).

.github/actions/preinstall-emulator/action.yml
Xexanos: Confirmed and fixed — you're right on both counts.

Reproduced the depth issue: `sdkmanager` sits at `cmdline-tools/latest/bin/sdkmanager`, which is depth 3 from `cmdline-tools` (`find`'s start dir is depth 0), so `-maxdepth 2` matched nothing, the not-found branch `exit 0`'d, and the retry loop never ran — the step was green but a no-op, which is exactly why no run ever demonstrated the retry firing.

Fix (commit `7d1fe52`): anchor to `cmdline-tools/latest/bin/sdkmanager` (the path android-emulator-runner itself expects), fall back to `find -maxdepth 3` only if `latest/` is absent, and log the resolved binary so a run visibly proves the install executes instead of skipping. That last part earned its keep immediately: with the step actually running, it exposed a latent second bug — `yes | sdkmanager --licenses` takes SIGPIPE (exit 141) under the composite's `set -o pipefail`, failing the step. Fixed in `e3c1823` by dropping pipefail around just that pipe (sdkmanager's own exit status still gates via `set -e`).

Both the resolution and the retry loop are now confirmed running in the green CI on this PR. Thanks for the catch.

PR #50, opened 2026-07-18
fix: load the whole library through cursor pagination (infinite scroll)

Xexanos:
## Why

Libraries with more than 50 books were silently truncated: `LibraryViewModel.load()` fetched one page of `GET /library/items` (contract default `limit` 50) and dropped `nextCursor`. Search results past 50 matches were cut off the same way.

## What

- **Infinite scroll**: a `derivedStateOf` scroll trigger requests the next page 8 rows before the end of the LazyColumn; pages are appended with the cursor from the previous response.
- **Stale-cursor safety**: follow-up pages are collected *inside* the search pipeline's `collectLatest` block, so changing the query cancels in-flight page loads and a stale cursor can never append to a newer query's results.
- **Footer states** (per ux-review findings): a `CircularProgressIndicator` with a TalkBack `contentDescription` while more pages exist; on a failed page load it flips to a tap-to-retry row (min 48dp touch target) instead of spinning forever — the loaded list is never replaced by the full-screen error. Scrolling back into the trigger zone also retries.
- **Previews + a11y checks**: new `LibraryLoadingMorePreview` / `LibraryLoadMoreFailedPreview`, both registered in `AccessibilityChecksTest`.
- **Strings**: two new strings in en and de.
- **Version bump 1.1.0 → 1.1.1** (versionCode 10101) with fastlane changelogs.

## Tests

- `loadMore follows the cursor and appends the next page` (incl. no request after the last page)
- `a failing next page keeps the list and flags a retryable error` (and retries successfully)
- `a new query restarts pagination from the first page`
- Existing debounce/error tests unchanged and green; `:app:lintDebug` clean.

## ux-review notes

Blocking findings (perpetual spinner after failure, missing a11y semantics) are addressed. Non-blocking leftovers, deliberately not in this PR: the design doc has no section for list-pagination states yet, and the screenshot-golden-test/lint anti-drift layer in CI is still open.

### DIFF
--- a/app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
@@ -15,11 +15,13 @@ import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
 import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.heightIn
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
 import androidx.compose.foundation.layout.width
 import androidx.compose.foundation.lazy.LazyColumn
 import androidx.compose.foundation.lazy.items
+import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.material.icons.Icons
 import androidx.compose.material.icons.automirrored.filled.MenuBook
 import androidx.compose.material.icons.filled.MusicNote
@@ -37,14 +39,18 @@ import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.runtime.derivedStateOf
 import androidx.compose.runtime.getValue
 import androidx.compose.runtime.mutableStateOf
+import androidx.compose.runtime.remember
 import androidx.compose.runtime.saveable.rememberSaveable
 import androidx.compose.runtime.setValue
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.semantics.contentDescription
+import androidx.compose.ui.semantics.semantics
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.text.style.TextAlign
 import androidx.compose.ui.text.style.TextOverflow
@@ -63,6 +69,7 @@ import io.github.xexanos.ratatoskr.ui.UiTestTags
 import io.github.xexanos.ratatoskr.ui.theme.RatatoskrTheme
 import io.github.xexanos.ratatoskr.ui.toMessage
 import kotlinx.coroutines.FlowPreview
+import kotlinx.coroutines.flow.MutableSharedFlow
 import kotlinx.coroutines.flow.MutableStateFlow
 import kotlinx.coroutines.flow.StateFlow
 import kotlinx.coroutines.flow.asStateFlow
@@ -75,6 +82,11 @@ import kotlin.math.roundToInt
 data class LibraryUiState(
     val loading: Boolean = false,
     val items: List<LibraryItemSummary> = emptyList(),
+    // Cursor for the page after the ones already in `items`; null once the last page is in.
+    val nextCursor: String? = null,
+    val loadingMore: Boolean = false,
+    // The last load-more attempt failed; the footer offers a tap-to-retry instead of a spinner.
+    val loadMoreError: Boolean = false,
     val error: String? = null,
 )
 
@@ -88,20 +100,34 @@ class LibraryViewModel(
 
     private val query = MutableStateFlow<String?>(null)
 
+    // Buffered so a request arriving while a page load is in flight is kept (and coalesced with
+    // any later ones) instead of dropped - the scroll position is still near the end when the
+    // page lands, and the UI's trigger won't fire again until the threshold is crossed anew.
+    private val loadMoreRequests = MutableSharedFlow<Unit>(extraBufferCapacity = 1)
+
     init {
         // One search pipeline for the whole screen: the initial (null) load runs immediately,
         // keystrokes are debounced, identical queries are ignored, and collectLatest cancels an
         // in-flight request when a newer query arrives so results can't land out of order.
+        // Follow-up pages are collected INSIDE the same block, so they always belong to the
+        // current query and die with it - a stale cursor can never append to fresh results.
         viewModelScope.launch {
             query
                 .debounce { if (it == null) 0L else SEARCH_DEBOUNCE_MS }
                 .distinctUntilChanged()
-                .collectLatest { load(it) }
+                .collectLatest { q ->
+                    load(q)
+                    loadMoreRequests.collect { loadNextPage(q) }
+                }
         }
     }
 
     fun search(query: String) {
         this.query.value = query.ifBlank { null }
+    }
+
+    fun loadMore() {
+        loadMoreRequests.tryEmit(Unit)
     }
 
     private suspend fun load(query: String?) {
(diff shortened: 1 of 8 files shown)


app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
```diff
@@ -112,11 +138,29 @@ class LibraryViewModel(
...
         }
     }
 
+    private suspend fun loadNextPage(query: String?) {
+        val cursor = _uiState.value.nextCursor ?: return
+        val client = connectionManager.client() ?: return
+        _uiState.value = _uiState.value.copy(loadingMore = true, loadMoreError = false)
+        _uiState.value = when (val result = client.listLibraryItems(query = query, cursor = cursor)) {
+            is ApiResult.Success -> _uiState.value.copy(
+                items = _uiState.value.items + result.data.items,
```
Xexanos: Appending `result.data.items` with no de-duplication against the already-loaded list means that if the server ever returns an id already present (cursor drift from a concurrent library mutation on the ABS side, or a server-side pagination bug), `items(state.items, key = { it.id })` (line 292) will throw `IllegalArgumentException: Key ... was already used` and crash the whole Library screen instead of degrading gracefully. Consider a defensive `distinctBy { it.id }` (or filtering out already-seen ids) when merging pages.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
```diff
@@ -219,13 +268,62 @@ private fun LibraryContent(
...
+                // the time the user reaches it. derivedStateOf keeps this from recomposing on
+                // every scroll frame: it only changes when the threshold is crossed.
+                val nearEnd by remember(listState) {
+                    derivedStateOf {
+                        val info = listState.layoutInfo
+                        val last = info.visibleItemsInfo.lastOrNull()?.index ?: -1
+                        last >= info.totalItemsCount - 1 - LOAD_MORE_THRESHOLD
+                    }
+                }
+                LaunchedEffect(nearEnd, state.nextCursor) {
```
Xexanos, line 293: `LaunchedEffect(nearEnd, state.nextCursor)` re-fires `onLoadMore()` automatically whenever a new page lands and the list is still shorter than `LOAD_MORE_THRESHOLD` rows — true even on first load for a small result set, no scroll required — and keeps re-firing on every subsequent page as long as `nextCursor` keeps changing. There's no cap on auto-chained pages and no backoff, so a server-side pagination bug that never returns a null cursor (or a very fine-grained page size on a large library) would make the client hammer `/library/items` continuously, back-to-back.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
```diff
@@ -75,6 +82,11 @@ import kotlin.math.roundToInt
 data class LibraryUiState(
     val loading: Boolean = false,
     val items: List<LibraryItemSummary> = emptyList(),
```
Xexanos, line 84: `items` only ever grows (`_uiState.value.items + result.data.items` on every page) with no upper bound or windowing, keeping every fetched item (and its Compose row state) in memory for the screen's lifetime instead of windowed loading (e.g. Paging 3). Likely fine for typical audiobook library sizes, but won't scale to unusually large libraries — worth a note if that's an accepted tradeoff for now.

app/src/test/java/io/github/xexanos/ratatoskr/ui/library/LibraryViewModelTest.kt
```diff
@@ -133,6 +133,114 @@ class LibraryViewModelTest {
...
+                        nextCursor = "next-after-$cursor",
+                    ),
+                )
+        }
+        val viewModel = LibraryViewModel(trustedConnectionManager())
+        settleState { viewModel.uiState.value.items.isNotEmpty() }
+        viewModel.loadMore()
+        settleState { viewModel.uiState.value.items.size == 2 }
+
+        viewModel.search("cat")
```
Xexanos, line 238: This test only calls `search("cat")` after `settleState { items.size == 2 }`, i.e. after the previous `loadMore()` has already fully completed — it doesn't cover the actual race the PR description calls out ("changing the query cancels in-flight page loads, so a stale cursor can never append to a newer query's results"). Consider a test that calls `search()` while a `loadNextPage` for the old query is still suspended waiting on the mock server, since that's the scenario the `collectLatest`-wrapped `loadMoreRequests.collect` structure is meant to protect against.

Xexanos (commented):
Automated review findings (see inline comments). Overall the design is sound (single search pipeline, `collectLatest` cancelling stale page loads, serialized `loadNextPage`), no domain logic leaked into the client. A few points worth a look before merging.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Fixed in c241d18: pages are now deduped by id before appending (`seen.add` filters out already-loaded ids), so a duplicate can never reach the LazyColumn's unique-key requirement.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Fixed in c241d18 with a no-forward-progress guard: if a page brings no new items after dedup, pagination stops (`nextCursor = null`). That covers the pathological non-null-cursor-forever case (and all-duplicate pages) without capping legitimate libraries. The auto-chaining to fill the viewport on a short list stays - that part is intended.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Accepted tradeoff for now (Variante 2 was chosen over Paging 3 deliberately). One correction: the list itself grows, but LazyColumn virtualizes, so Compose row state is NOT retained for offscreen items. Rough estimate: ~0.3-0.5 KB/item text-only, so OOM territory is ~50k-100k items on a low-heap device - not a concern for realistic audiobook libraries. Captured with the full analysis in #65 (incl. the O(n^2) append as a cheap independent win).

app/src/test/java/io/github/xexanos/ratatoskr/ui/library/LibraryViewModelTest.kt
Xexanos: Good catch - fixed in c241d18. Added `a query change cancels an in-flight page load so its stale page never appends`: the next-page response is held on a CountDownLatch so `search()` arrives while loadNextPage is genuinely suspended, then asserts the stale page never lands after the latch releases.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Confirmed — thanks for the quick fix, that closes the crash risk.

app/src/test/java/io/github/xexanos/ratatoskr/ui/library/LibraryViewModelTest.kt
Xexanos: The CountDownLatch-based repro is exactly what was missing — confirmed it covers the race. Thanks.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Good correction, thanks — I underestimated how aggressively LazyColumn disposes offscreen composables. The 50k-100k floor with #65 tracking the O(n^2) append as a follow-up sounds like the right call for now.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Makes sense for the fully-duplicate-page case. One follow-up: what happens on a page that's a *partial* duplicate — say the server returns 20 items where 19 were already loaded and 1 is new? That still counts as forward progress under a `seen.add`-based check, so pagination keeps going. Is that an intentional "slow drain is fine, only a true stall is a bug" stance, or worth an additional safeguard (e.g. a min-new-items-per-page threshold) for a badly-behaved cursor that repeats itself with slowly-changing content?

Xexanos: Agreed — leaving it as-is, no min-new-items threshold.

The partial-duplicate case is already bounded and non-fatal: dedup prevents both the crash and duplicate bloat, every page adds ≥1 new item so the list grows monotonically and pagination terminates in ≤N pages — worst case just extra requests against a misbehaving server, never an infinite loop or corrupted state (the true-stall case is caught by `fresh.isEmpty()`).

A min-new threshold would be the wrong fix: '1 new item' *is* genuine forward progress, so any threshold >0 risks silently truncating a legitimate library when a real page happens to have few new items (cursor overlap at page boundaries, small page size, server-side filtering) — i.e. reintroducing exactly the #50 'not all books shown' bug, but hidden and hard to diagnose. Trading a low-probability, non-fatal, server-caused inefficiency for a high-impact silent-data-loss risk is a bad deal.

A persistent 19/20-overlap stream is a server/ABS cursor-correctness bug (or pagination over a concurrently-mutating collection, which isn't cleanly solvable client-side and needs a stable/snapshot cursor server-side) — it belongs fixed there, not papered over in the client. If we ever want a safety net it should be **diagnostic** (a debug log when a page is mostly duplicates), never behavioral: surface the server bug without ever dropping data.

Stance, for the record: any forward progress continues; deliberately no min-new threshold, to avoid silent truncation.

Issue #72, opened 2026-07-19
Implement the woven-knot loader for felt waits (first library load, sign-in, connect)


Xexanos:
Surfaced during the #68 ux-review.

## Problem

The UX design (docs/ux-design.html, Motion & loading §1099-1131, rule 1 §1127) mandates a **woven-knot loader** for the three *felt* waits — notably the **first library load** — while short inline waits (search, pull-to-refresh) keep the standard M3 `CircularProgressIndicator`. The knot loader is **not implemented anywhere** in `app/src/main`; those felt-wait paths currently use a plain `CircularProgressIndicator` (e.g. the Library full-screen `state.loading` branch).

#68 (reload) routes a new path through it — retry-with-nothing-shown reloads via the full-screen spinner — so reload now leans on this gap further, though it doesn't introduce it.

## Scope

- Build the knot-loader composable per the design (and honor the reduced-motion rule, §1129).
- Use it for the felt waits (first library load; likely the connect/sign-in waits too), leaving pull-to-refresh and other short inline waits on the standard indicator.

Deferred; pre-existing, app-wide.

Issue #71, opened 2026-07-19
Error messages are hardcoded English in ViewModels; error states lack guidance copy


Xexanos:
Surfaced during the #68 ux-review.

## Problem

Error text shown in the UI is produced in ViewModels (no `Context`), so it is **hardcoded English**, not localized:
- `LibraryViewModel` sets `LibraryUiState(error = "No server configured.")` (LibraryScreen.kt), and maps API failures via `error.toMessage()` — both run in the VM, off the Composition, so neither can use `stringResource`.
- Same pattern in `SignInScreen.kt` and `SpeakersScreen.kt` (`"No server configured."`).

This violates the copy convention (UI strings live in `strings.xml`, English source, localized to de). Now more prominent because #68 gave the Library error state a first-class "Try again" affordance.

## Also (ux-design ERR/EMPTY, §1087-1095)

The full-screen error state renders just the message + button. The design wants error/empty states to **name whose state it mirrors** (server / network / search) and pair a title with guidance ("…Try a different title or author"). Whether the current `toMessage()` output meets that depends on its copy — worth auditing.

## Direction

Model errors as a resource id (or a sealed error type) carried in UI state and resolved in the Composable, rather than a pre-formatted String from the VM. Then give each error state a title + guidance line, localized. Applies to Library, Sign-in, Speakers.

Deferred; not part of #68 (pre-existing).

PR #73, opened 2026-07-19
feat: reload the library (pull-to-refresh + retry on load error)

### LINKED ISSUE #68: Library: reload the list (pull-to-refresh + retry on load error)
## What

There is currently no way to reload the library list. If the initial load fails, the Library
screen shows only the full-screen error text with **no retry** (the `state.error != null` branch in
`LibraryScreen.kt`) — the only way out is to navigate away. The load-more failure already has a
tap-to-retry footer (from #50), but the full-screen error and the general "refresh the list" case
do not.

## Scope

- **Pull-to-refresh** (primary): swipe-down reload of the current view (respects the active search
  query), Material 3 `PullToRefreshBox`.
- **Retry on the full-screen error state**: a "Try again" affordance so a failed initial load is
  recoverable in place. (Pre-existing gap — `load()` could always fail; not introduced by #50.)
- Optional: **auto-reload on return to the screen** (lifecycle `ON_RESUME`) so the list is fresh
  after a long absence. Could ship here or separately.

## Mechanism

The search pipeline in `LibraryViewModel` uses `distinctUntilChanged()`, so re-emitting the same
query does not reload. A manual reload needs an explicit refresh trigger (e.g. a
`MutableSharedFlow<Unit>`) merged into the pipeline that runs `load(query.value)` **without** the
search debounce/distinct, reusing the existing `collectLatest` so it cancels any in-flight page
load and keeps the current query. Pull-to-refresh and error-retry share this one `refresh()` path.
Needs unit tests (refresh keeps the query; cancels in-flight `loadNextPage`; clears/repopulates
state correctly).

## Notes

- Deliberately **not** part of #50 (that PR is the pagination bugfix and is review-complete); this
  is a separate feature PR to land after #50 merges.
- New interaction → should go through the `ux-review` and likely a short addition to the Library
  section (screen 03) of `docs/ux-design.html` (pull-to-refresh + error-retry states).

Xexanos:
Closes #68.

## What

- **Pull-to-refresh** on the loaded list (Material 3 `PullToRefreshBox`, standard indicator per the design's motion rule for short inline waits). Reloads the current query in place.
- **Retry on the full-screen error state** — a "Try again" button (fills the pre-existing gap where a failed initial load was unrecoverable).
- **Failed refresh over an existing list** keeps the list and offers **retry as a snackbar action** (not the full-screen error) — retry is always the first offered action.
- ViewModel: one `collectLatest` pipeline funnels query changes *and* explicit refreshes; a refresh reloads the current query, bypassing debounce/`distinctUntilChanged`, and cancels any in-flight load or page fetch.
- New `library_retry` (en + de), `LibraryErrorPreview` registered in the a11y checks, and a **Library reload decision added to `docs/ux-design.html`** so the design stays the contract.
- Version 1.1.1 → 1.2.0 + changelogs.

## Tests

Three new `LibraryViewModelTest` cases: refresh reloads the current query in place; retry after a failed load re-issues it and clears the error; a failed refresh keeps the list and surfaces a one-shot `refreshError`. Existing suite unchanged. `assembleDebug` + `testDebugUnitTest` + `lintDebug` green.

## ux-review

Ran the ux-review agent. Its one blocking finding on this work — the refresh-failure snackbar offered no retry action (violates "retry is always offered") — is **fixed** (snackbar now has a Try-again action). Design-doc line added per its Q-1.

Deferred to their own tickets (pre-existing, out of scope here):
- **#71** — error messages hardcoded in ViewModels (no `Context`), and error states lacking guidance copy.
- **#72** — the woven-knot loader for felt waits is unimplemented app-wide; the full-screen load path still uses a plain spinner.

Still open (program-level, noted before): no screenshot-golden-test / detekt anti-drift layer in CI, and the instrumented a11y suite isn't a required check.

### DIFF
--- a/app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
@@ -29,14 +29,21 @@ import androidx.compose.material.icons.filled.PlayArrow
 import androidx.compose.material.icons.filled.Search
 import androidx.compose.material.icons.filled.SearchOff
 import androidx.compose.material.icons.filled.Settings
+import androidx.compose.material3.Button
 import androidx.compose.material3.CircularProgressIndicator
+import androidx.compose.material3.ExperimentalMaterial3Api
 import androidx.compose.material3.Icon
 import androidx.compose.material3.IconButton
 import androidx.compose.material3.LinearProgressIndicator
 import androidx.compose.material3.MaterialTheme
 import androidx.compose.material3.OutlinedTextField
+import androidx.compose.material3.Scaffold
+import androidx.compose.material3.SnackbarHost
+import androidx.compose.material3.SnackbarHostState
+import androidx.compose.material3.SnackbarResult
 import androidx.compose.material3.Surface
 import androidx.compose.material3.Text
+import androidx.compose.material3.pulltorefresh.PullToRefreshBox
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.LaunchedEffect
 import androidx.compose.runtime.derivedStateOf
@@ -76,6 +83,8 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.flow.collectLatest
 import kotlinx.coroutines.flow.debounce
 import kotlinx.coroutines.flow.distinctUntilChanged
+import kotlinx.coroutines.flow.map
+import kotlinx.coroutines.flow.merge
 import kotlinx.coroutines.launch
 import kotlin.math.roundToInt
 
@@ -87,8 +96,16 @@ data class LibraryUiState(
     val loadingMore: Boolean = false,
     // The last load-more attempt failed; the footer offers a tap-to-retry instead of a spinner.
     val loadMoreError: Boolean = false,
+    // A pull-to-refresh is reloading over an existing list (drives the pull indicator).
+    val refreshing: Boolean = false,
+    // One-shot message for a refresh that failed while a list was already shown (a snackbar).
+    val refreshError: String? = null,
     val error: String? = null,
 )
+
+// A single reload of the library. `userTriggered` distinguishes a pull-to-refresh / retry (which
+// reloads over an existing list without the full-screen spinner) from a query change.
+private data class Reload(val query: String?, val userTriggered: Boolean)
 
 @OptIn(FlowPreview::class)
 class LibraryViewModel(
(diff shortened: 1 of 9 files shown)


app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
```diff
@@ -193,34 +241,56 @@ fun LibraryScreen(
...
+            val result = snackbarHostState.showSnackbar(
+                message = refreshError,
+                actionLabel = retryLabel,
+            )
+            viewModel.clearRefreshError()
+            if (result == SnackbarResult.ActionPerformed) viewModel.refresh()
+        }
+    }
+
+    Scaffold(snackbarHost = { SnackbarHost(snackbarHostState) }) { padding ->
```
Xexanos: Nesting a second `Scaffold` here may double-apply system-bar insets. `MainActivity` already wraps the whole app in one top-level `Scaffold` under `enableEdgeToEdge()`, and applies its `innerPadding` via a plain `Modifier.padding(innerPadding)` on a `Surface` — that doesn't mark window insets as consumed for descendants. This inner `Scaffold` uses the default `contentWindowInsets` (`WindowInsets.systemBars`), so it will likely compute status-bar/nav-bar padding a second time on top of what the outer Scaffold already applied. No other screen in the codebase sets `contentWindowInsets`, so Library would be the first place this surfaces — worth comparing the top/bottom gap against e.g. `NowPlayingScreen` on-device before merging. If it does double up, either drop the nested `Scaffold` (host the `SnackbarHost` in a `Box` instead) or pass `contentWindowInsets = WindowInsets(0)` here since the outer Scaffold already owns the insets.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
```diff
@@ -130,8 +156,23 @@ class LibraryViewModel(
...
+        // A user-triggered reload OVER an existing list shows the pull indicator and keeps the
+        // list; everything else (initial load, query change, retry with nothing shown) uses the
+        // full-screen spinner.
+        val showAsRefresh = userTriggered && _uiState.value.items.isNotEmpty()
+        _uiState.value =
+            if (showAsRefresh) _uiState.value.copy(refreshing = true, refreshError = null)
+            else _uiState.value.copy(loading = true, error = null)
         val client = connectionManager.client()
         if (client == null) {
             _uiState.value = LibraryUiState(error = "No server configured.")
```
Xexanos: This branch unconditionally resets to `LibraryUiState(error = "No server configured.")`, regardless of `showAsRefresh` — unlike the `ApiResult.Failure` branch below, which keeps the list and reports via `refreshError` when `showAsRefresh` is true. So a refresh that fails because the server became unconfigured mid-session wipes the already-loaded list and drops to the full-screen error, contradicting the PR's own contract ("a failed refresh over an existing list keeps the list"). Narrow edge case, but for consistency:
```kotlin
if (client == null) {
    _uiState.value = if (showAsRefresh) {
        _uiState.value.copy(refreshing = false, refreshError = "No server configured.")
    } else {
        LibraryUiState(error = "No server configured.")
    }
    return
}
```
No test currently covers this path either.

Xexanos (commented):
Automated review pass (Claude Code). Two findings inline below — one worth verifying on-device before merge, one a narrow edge-case inconsistency. Nothing blocking; full write-up was also shared in chat.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Fixed in 241f4f2 — dropped the nested Scaffold entirely (no other screen uses one; MainActivity owns the single top-level Scaffold). The snackbar now lives in a Box overlay (`SnackbarHost` aligned bottom), which sits inside the already-inset content area, so the double system-bar padding is structurally gone rather than papered over with `contentWindowInsets(0)`.

app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
Xexanos: Good catch — fixed in 241f4f2. The `client == null` branch now honours `showAsRefresh` exactly like the failure branch: a refresh over an existing list keeps the list and sets `refreshError` instead of wiping to the full-screen error. Added a test (`a refresh that fails because the server became unconfigured keeps the list`) using `FakeConnectionStore` cleared mid-session — the in-memory fake avoids the real DataStore's Windows file-rename flakiness.

PR #75, opened 2026-07-19
feat: replace the loading spinner with the animated knot loader

### LINKED ISSUE #72: Implement the woven-knot loader for felt waits (first library load, sign-in, connect)

Xexanos:
Closes #72.

Brings the woven-knot loader from the UX design to the app, replacing the generic `CircularProgressIndicator` on the app's full/section waits.

## What
- **`ui/KnotLoader.kt`** — the ash-green knot from the logo holds still while a copper squirrel runs the strand centre line, trailing a fading comet. Redrawn in a `Canvas` with `PathMeasure` (Android can't run the SMIL reference SVG); **no new dependency, no server**. Path data lifted from `docs/logo/ratatoskr-spinner.svg`.
- **Time threshold** — every indeterminate full/section wait renders nothing for ~500 ms and only then escalates to the knot (`rememberDelayedVisible`), so quick waits never flash it. Wired into MainActivity (cold start), Connect (cert inspect), Speakers (loading/starting), Library (load/search), Now playing.
- **Kept inline** as standard M3 spinners: the in-button sign-in spinner and the library pagination footer (a full knot doesn't belong in a button or a list footer).
- **Reduced motion** — `ANIMATOR_DURATION_SCALE = 0` via new `LocalReducedMotion` renders the static knot; a text label then carries the wait, falling back to "Loading…" when the call site gives none (design rule 3).
- **Design doc** — the motion rule is updated from the "felt waits only" site list to the time threshold. This scope change (drop the per-screen classification, gate everything behind the threshold) was deliberate and agreed during planning.

## Verification
- `:app:testDebugUnitTest` (incl. new `KnotLoaderTest` — pure seam/wrap + reduced-motion math), `:app:lintDebug`, `:app:assembleDebug` green.
- `:app:connectedDebugAndroidTest` `AccessibilityChecksTest` (21 tests incl. two new knot previews) green on the emulator.
- Visually confirmed on the emulator at several animation frames: light + dark themes, the seam crossing (wrap-split draws cleanly), and the static reduced-motion frame.

## Notes / not in scope
- A `ux-review` pass flagged that there's still no **screenshot golden test / lint in CI** — the known open tooling gap; a Canvas animation is exactly what a golden would pin. Tracked separately, not addressed here.
- `rememberReducedMotion()` snapshots the setting once (won't react to a live toggle without recomposition) — acceptable for v1.

Version: 1.2.1 → **1.3.0** (feat, minor).


app/src/androidTest/java/io/github/xexanos/ratatoskr/ui/AccessibilityChecksTest.kt
```diff
@@ -92,6 +92,8 @@ class AccessibilityChecksTest {
     @Test fun nowPlayingPaused() = runChecks { NowPlayingPausedPreview() }
     @Test fun nowPlayingEmpty() = runChecks { NowPlayingEmptyPreview() }
     @Test fun settings() = runChecks { SettingsPreview() }
+    @Test fun knotLoader() = runChecks { KnotLoaderPreview() }
```
Xexanos, line 126: **Risk: `speakersLoading()`/`libraryLoading()` (line 90/84) likely now test an empty box instead of the loader.**

`SpeakersLoadingPreview`/`LibraryLoadingPreview` render with `loading = true` as a constant; the loader now sits behind `rememberDelayedVisible(state.loading)` (a 500ms `delay()` inside a `LaunchedEffect`). `runChecks` calls `compose.setContent(...)` and immediately `tryPerformAccessibilityChecks()`, without advancing the compose clock. Compose's idle detection doesn't wait out a plain coroutine `delay()` (it only tracks pending recomposition/layout/draw and the animation clock) — so the check most likely runs before `visible` flips to `true`, and ends up checking an empty `Box` instead of loader + label.

The tests still stay green (nothing in an empty box can violate an a11y rule), but they lose their actual coverage of the real screens' loading state — the PR description's "21 tests green" reads as if coverage held. `knotLoader()`/`knotLoaderReduced()` (here) do exercise `KnotLoader` itself directly (ungated), but that doesn't cover whether `SpeakersContent`/`LibraryContent` still expose something accessible in their real loading state.

Suggestion: advance the compose clock before the check in these two tests (e.g. `compose.mainClock.advanceTimeBy(600)`), or give `rememberDelayedVisible` a `delayMillis` override usable from tests/previews. Worth verifying on the emulator — this is inferred from the diff, not actually run.

app/src/main/java/io/github/xexanos/ratatoskr/ui/KnotLoader.kt
```diff
@@ -0,0 +1,332 @@
...
+        Box(Modifier.fillMaxSize(), contentAlignment = Alignment.Center) {
+            KnotLoader(label = stringResource(R.string.app_loading))
+        }
+    }
+}
+
+@Preview(name = "Knot loader - reduced motion", widthDp = 360, heightDp = 300)
+@Composable
+internal fun KnotLoaderReducedPreview() = RatatoskrTheme {
+    androidx.compose.runtime.CompositionLocalProvider(LocalReducedMotion provides true) {
```
Xexanos: Minor: `androidx.compose.runtime.CompositionLocalProvider` is fully-qualified here instead of imported — inconsistent with the otherwise clean import list in this file.

app/src/main/java/io/github/xexanos/ratatoskr/ui/KnotLoader.kt
```diff
@@ -0,0 +1,332 @@
...
+ * runner itself (a translucent halo under a solid dot). [head] is the head position as a fraction
+ * `[0,1)` of the path length.
+ */
+private fun DrawScope.drawComet(geo: KnotGeometry, head: Float, color: Color) {
+    val length = geo.length
+    if (length <= 0f) return
+    val headDist = head * length
+    val tailLen = length * TAIL_FRACTION
+    val seg = tailLen / TAIL_SEGMENTS
+    val scratch = android.graphics.Path()
```
Xexanos: Minor/perf: `drawComet` allocates a new `android.graphics.Path()` and a `FloatArray(2)` on every draw call (i.e. every animation frame while the loader is visible), and `wrappedRanges` allocates a new `List<Pair<Float,Float>>` for each of the 24 tail segments. Probably fine since the loader only appears on genuinely slow waits — if it shows up in profiling, a reused scratch object (remember-scoped) would be the next step.

Xexanos (commented):
## Overview

Replaces `CircularProgressIndicator` with a brand-specific "knot loader" (`KnotLoader.kt`) across five call sites (cold start, cert inspection, speakers, library, now playing). The knot is a static `Canvas`-drawn path (parsed from the logo SVG's `d` data via `PathMeasure`) with an animated copper "runner" comet running the centerline. A new `rememberDelayedVisible(active, delayMillis = 500)` gates all of these behind a 500ms threshold so fast waits never flash the loader, and `LocalReducedMotion` (reading `ANIMATOR_DURATION_SCALE`) freezes the animation and falls back to a text label for accessibility. The design doc, SVG reference, and logo generator are updated in lockstep (animation duration 2.4s -> 4s in all three places - consistent). Solid PR: pure helper functions (`wrappedRanges`, `reducedMotionFromScale`) are unit-tested, and a new `KnotLoaderPreview`/`KnotLoaderReducedPreview` pair is wired into the existing instrumented accessibility suite.

## Findings

See the inline comments on this review for the specifics:

1. **Risk** - the 500ms delay gate likely defeats `speakersLoading()`/`libraryLoading()` in `AccessibilityChecksTest`, silently turning them into checks against an empty box rather than the loader + label.
2. **Minor** - fully-qualified `CompositionLocalProvider` instead of an import in `KnotLoaderReducedPreview`.
3. **Minor/perf** - per-frame allocations in `drawComet`/`wrappedRanges`; likely fine given how rarely the loader shows, flagged for awareness.

## What's good

- The delay-gate design (`rememberDelayedVisible`) is correctly keyed on `active` - flipping to `false` resets immediately, flipping to `true` restarts the timer via `LaunchedEffect(active)` cancellation.
- `wrappedRanges` cleanly isolates the one genuinely tricky bit of math (seam-straddling on a closed `PathMeasure` path) into a pure, well-tested function.
- Reduced-motion handling is layered correctly: static render + mandatory text label fallback, single `LocalReducedMotion` read at the theme root.
- Design doc, SVG reference, and the Kotlin generator were all updated together for the 2.4s -> 4s animation duration change - no drift between the three.
- Follows the app's working agreement: no new dependency, no server involvement, path data lifted from the existing SVG asset.

No security concerns and no scope creep beyond `Closes #72`.

app/src/androidTest/java/io/github/xexanos/ratatoskr/ui/AccessibilityChecksTest.kt
Xexanos: Good catch, exactly right - both tests were checking the empty box shown before the 500ms gate, so they covered nothing. Verified on the emulator: the new `runChecksAfterLoaderDelay` pauses `autoAdvance`, steps past the gate with `advanceTimeBy(600)`, and **then asserts** the loader is actually present (`knotShown || labelShown`, robust to reduced-motion in case CI disables animations) before running the a11y checks. All 21 tests green locally, including the now genuinely-covered loading states. (ce9695f73807ad88a580e8cfcd0acc960bc6feb2)

app/src/main/java/io/github/xexanos/ratatoskr/ui/KnotLoader.kt
Xexanos: Done - `CompositionLocalProvider` is now imported and the preview uses the short name. (ce9695f73807ad88a580e8cfcd0acc960bc6feb2)

app/src/main/java/io/github/xexanos/ratatoskr/ui/KnotLoader.kt
Xexanos: The cheap hot-path allocations are gone: `scratch` Path + `pos` FloatArray are now remember-scoped and reused across frames, and the per-segment `asComposePath()` wrapper is dropped (the reused Compose path is drawn directly). I deliberately kept the small list from `wrappedRanges` - it keeps the seam logic as one pure, unit-tested function, and the loader is only briefly visible on genuinely slow waits anyway (noted in a comment). (ce9695f73807ad88a580e8cfcd0acc960bc6feb2)

Issue #78, opened 2026-07-19
Reconsider initials-based cover placeholders (alphabetical list makes them repetitive)


Xexanos:
The library placeholder/fallback tile shows the title's first letter (see `CoverThumb` in `LibraryScreen.kt`, `CoverArt` in `NowPlayingScreen.kt`; design rationale in docs/ux-design.html — "initials on a brand gradient").

In practice the library list is sorted alphabetically, so any stretch of the list without cover art shows a column of tiles with the **same letter** — the initial carries no distinguishing information exactly where it is meant to help telling rows apart.

With #55 landing, real covers will dominate and the placeholder becomes a loading/fallback state rather than the primary look, so this is not urgent — but the decision deserves a re-check:

- Keep initials (cheap, looks "designed", matches ux-design.html)?
- Vary the tile another way (per-title gradient/color hash, so identical letters still differ)?
- Drop the letter for a neutral book icon?

Decide and either update docs/ux-design.html or the implementation to match.

_Spun out of the #55 design session (2026-07-19)._

Xexanos:
Two judgment notes from the #55 UX review (PR #81); the second has since moved to its own issue:

1. **The empty-title fallback icon drifts from the design.** When a title has no usable initial, `CoverImage` (formerly `CoverThumb`/`CoverArt`) falls back to `Icons.Default.MusicNote`. docs/ux-design.html (Iconography > Placeholders) specifies "the title's initial on a brand gradient, with a tinted `MenuBook`/`LibraryMusic` icon as fallback - not a music-note glyph". Whatever this issue decides about the initials, the icon choice should be settled with it - it is the same placeholder surface (`app/src/main/java/io/github/xexanos/ratatoskr/ui/common/CoverImage.kt`, `InitialTile`).

2. ~~The now-playing cover radius (20dp vs designed 24-28dp)~~ - split out to #85: it is a theme-shape-token decision independent of the placeholder question.

For context: PR #81 already aligned the two drifts that were unambiguous - cover thumbnails now default to the designed 8dp (`shapes.small`), and the cover crossfade honors reduced motion.

Xexanos:
**Decision** (design session 2026-07-20): the initials go. The placeholder splits into two quiet tiles, and the no-cover state shows the **tinted knot mark** — the Ratatoskr logo's frame in monochrome `onSecondaryContainer` on the `secondaryContainer` surface.

### Reasoning

[...]

Issue #52, opened 2026-07-18
Library: pin actively-listened books to the top and mark them visually


Xexanos:
## What

Books the user is actively listening to (progress > 0, not finished) should lead the library list and carry a distinct visual marker — the equivalent of ABS's "Continue Listening" shelf, adapted to the app's single-list layout.

## Notes

- `LibraryItemSummary.progress` already carries per-item progress, and rows already render a progress bar; what's missing is **ordering** and a **dedicated affordance** (e.g. a pinned "Continue listening" section or a badge on the row).
- Ordering interacts with the cursor pagination from #50: with infinite scroll the client only ever sees the loaded pages, so "in-progress first" must be **server-side ordering** (contract change) or a small dedicated "continue listening" endpoint/section fetched separately — client-side sorting of partial pages would be wrong.
- Search results should probably stay relevance/alphabetical — the pinning applies to the unfiltered library view.
- UX: needs a section in docs/ux-design.html for screen 03 (Library) before implementation; orient on the official ABS app's Continue Listening presentation.

Xexanos:
## Decision record (grilling session, 2026-07-20) — app side, buildable

Server side is done (v1.4.0, `GET /library/in-progress` → `LibraryItemList`, no cursor, `limit` ≤ 50 default 25). Domain terms recorded in `CONTEXT.md` (Browse list / Continue-listening shelf / In progress).

| # | Decision |
|---|---|
| 1 | **Vertical shelf section** above the browse list, same `LibraryRow`s, one LazyColumn — no carousel, no in-list pinning (the endpoint shape already settled this server-side) |
| 2 | **No deduplication** — the shelf is a view, not a partition; shelf books keep their place in the browse list. LazyColumn keys namespaced (`shelf:{id}`) |
| 3 | **Search hides the shelf immediately** (UI-side, non-blank query); results stay alphabetical; clearing the search restores the shelf from held state, no refetch |
| 4 | **Refresh:** initial load and pull-to-refresh fetch shelf + first page in parallel; plus a silent shelf refetch on every Library screen re-entry. Shelf fetches only when the query is blank. Event-coupled refresh (session ends while on Library) lands with the mini player, #79 |
[...]

PR #81, opened 2026-07-19
Cover art in library and now-playing

### LINKED ISSUE #55: Cover art in library and now-playing
The library projection carries no cover URL today. Once the server serves cover images (server SPEC section 16), show artwork in the library and now-playing views. Blocked on the server contract adding a cover URL to the library item projection.

_Deferred post-v1 feature, moved out of app SPEC section 14._

Xexanos:
Closes #55.

## What

The library rows and the now-playing screen now show real cover art, served scaled by the Ratatoskr server's cover proxy (contract 1.3.1, server v1.4.0). Books without a cover keep the title-initial tile, and Settings gains a "Clear image cache" action.

## How

- **Contract bump to 1.3.1** and wrapper absorption: `coverUrl` is origin-relative on the wire since 1.3.x; the mappers resolve it against the client's base URL so the domain model keeps carrying an absolute, bearer-authenticated URL (absolute values pass through — tolerant reader).
- **`coversCallFactory` on `RatatoskrClient`**: shares the main client's TOFU trust, bearer/refresh auth and connection pool, but owns a separate dispatcher so a scroll burst of cover loads can never queue playback commands or the session poll (OkHttp admission is per-dispatcher, same host).
- **Coil 3** (Apache-2.0, no Google dependencies, F-Droid-safe), app module only. One app-lifetime `ImageLoader` that follows server/certificate changes through a per-request delegate; provided via a CompositionLocal so tests inject a `FakeImageLoaderEngine`.
- **Bucketed `?h=`**: a Coil interceptor quantizes the measured layout height to {128, 256, 512, 1024} (rounded up, capped at 1024 — deliberate deviation from the contract's "omit for full size" hint; an unclamped original can be 3000×3000 for no visible gain). Call sites carry no size constants; same-bucket surfaces share one cached image.
- **Caching**: 50 MB disk LRU under `cacheDir/covers`. HTTP cache headers are deliberately ignored — verified against the ABS source: its cover endpoint sends no Cache-Control/ETag/Last-Modified on the proxy's request path, so respecting them would mean never caching. Stale covers are the accepted trade-off, bounded by LRU, the new Settings action, and the cache wipe on sign-out / forget-server (a signed-out device keeps no artwork of the library it left). Server-side follow-ups: Xexanos/ratatoskr-server#98 (null coverUrl when no cover exists), plus a server-owned Cache-Control policy.
- **One shared `CoverImage` composable** (`ui/common`) used by the library and now-playing — the continue-listening shelf (#52, builds on this branch) reuses it as-is. Placeholder and error state are identical by design; the placeholder's visual design is under review in #78.

## Tests

- Unit: coverUrl resolution (relative/absolute/null), bucket boundaries, settings ViewModel incl. cache-clear flows.
- Compose (instrumented): the tile's three states against a fake engine — success replaces the initials tile, failure keeps it, null URL issues no request.
- Integration flow: the cover request carries the bearer token and a bucketed `?h=` through the TOFU-pinned stack, and the decoded PNG replaces the placeholder.
- R8 gates pass (`assembleRelease`, minified androidTest); release APK 1.8 → 1.91 MB. The minified test APK needed one new `-dontwarn` (coil-test's transitive material references appcompat) and a narrow `coil3.ColorImage` keep — kept narrow on purpose so the minified run still validates Coil under the release keep rules.

## Docs

SPEC §12 records Coil 3 with the selection rationale and caching policy; §13 gains `ui/common` and `covers/`; CONTEXT.md defines Cover, Cover URL, Cover bucket and Placeholder tile. Version 1.4.0 (+ fastlane changelogs en-US/de-DE).

### DIFF
--- a/app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
@@ -34,10 +34,12 @@ internal class CoverHeightInterceptor : Interceptor {
     }
 
     /**
-     * Only the cover proxy understands `h`; leave any other image URL alone. Matched on the
-     * proxy's path shape rather than a bare `/cover` suffix, so a pass-through absolute URL
-     * from a foreign origin (the wrapper's tolerant-reader case) is not decorated with a
-     * parameter only Ratatoskr defines.
+     * Only the cover proxy understands `h`; leave any other image URL alone. A heuristic on
+     * the proxy's path shape - deliberately host-agnostic, because knowing the server origin
+     * is the wrapper's concern, not this interceptor's. If the server ever returns to sending
+     * absolute URLs (the wrapper's tolerant-reader case), its own covers keep matching; a
+     * foreign URL that happens to share the exact path shape would get one extra query
+     * parameter, which servers ignore.
      */
     private fun isCoverUrl(url: String): Boolean {
         val path = url.substringBefore('?')
--- a/app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
@@ -1,0 +1,39 @@
+/*
+ * Ratatoskr Android app
+ * Copyright (C) 2026  Ratatoskr contributors
+ * SPDX-License-Identifier: GPL-3.0-or-later
+ */
+package io.github.xexanos.ratatoskr.covers
+
+import coil3.intercept.Interceptor
+import coil3.request.ImageResult
+import coil3.size.Dimension
+
+/**
+ * Appends the bucketed `?h=` parameter to cover requests from the size Coil resolved out of
+ * the layout, so call sites stay dumb: every surface renders `CoverImage` and the right
+ * server-side scale falls out of the composable's measured height (list row, shelf,
+ * now-playing - no per-screen size constants that could drift). An unresolved height
+ * (Dimension.Undefined) leaves the URL untouched, which the server serves as full size.
+ *
+ * Runs before Coil's engine, so the rewritten URL is also the memory/disk cache key: two
+ * surfaces whose heights land in the same bucket share one cached image.
+ */
+internal class CoverHeightInterceptor : Interceptor {
+
+    override suspend fun intercept(chain: Interceptor.Chain): ImageResult {
+        val data = chain.request.data
+        val height = chain.size.height
+        if (data !is String || !isCoverUrl(data) || height !is Dimension.Pixels) {
+            return chain.proceed()
+        }
+        val request = chain.request.newBuilder()
+            .data(appendCoverHeightParam(data, height.px))
+            .build()
+        return chain.withRequest(request).proceed()
+    }
+
+    /** Only the cover proxy understands `h`; leave any other image URL alone. */
+    private fun isCoverUrl(url: String): Boolean =
+        url.substringBefore('?').endsWith("/cover")
+}
--- a/app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
@@ -33,7 +33,14 @@ internal class CoverHeightInterceptor : Interceptor {
         return chain.withRequest(request).proceed()
     }
 
-    /** Only the cover proxy understands `h`; leave any other image URL alone. */
-    private fun isCoverUrl(url: String): Boolean =
-        url.substringBefore('?').endsWith("/cover")
+    /**
+     * Only the cover proxy understands `h`; leave any other image URL alone. Matched on the
+     * proxy's path shape rather than a bare `/cover` suffix, so a pass-through absolute URL
+     * from a foreign origin (the wrapper's tolerant-reader case) is not decorated with a
+     * parameter only Ratatoskr defines.
+     */
+    private fun isCoverUrl(url: String): Boolean {
+        val path = url.substringBefore('?')
+        return path.endsWith("/cover") && path.contains("/library/items/")
+    }
 }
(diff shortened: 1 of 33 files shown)


Xexanos (commented):
Automated review pass — inline notes below, nothing blocking.

app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
```diff
@@ -0,0 +1,46 @@
...
+
+    /**
+     * Only the cover proxy understands `h`; leave any other image URL alone. Matched on the
+     * proxy's path shape rather than a bare `/cover` suffix, so a pass-through absolute URL
+     * from a foreign origin (the wrapper's tolerant-reader case) is not decorated with a
+     * parameter only Ratatoskr defines.
+     */
+    private fun isCoverUrl(url: String): Boolean {
+        val path = url.substringBefore('?')
+        return path.endsWith("/cover") && path.contains("/library/items/")
```
Xexanos, line 46: isCoverUrl only checks the path shape, not the host. The doc comment above (lines 37-40) says a pass-through absolute URL from a foreign origin won't get decorated with `?h=` — but as written, any absolute URL whose path happens to end in `/cover` and contain `/library/items/` gets the bucket param appended regardless of host. In practice this only bites if a future server reverts to sending absolute cover URLs (same origin, same path shape), so the real-world risk is low — but either tighten this to also compare against the client's known origin, or soften the comment so it doesn't overstate what the check does.

app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverImages.kt
```diff
@@ -0,0 +1,77 @@
...
+    // The fetcher factory memoizes its callFactory lambda, so hand it one stable delegating
+    // factory that re-reads the current client per call - this is what makes a server or
+    // certificate change take effect without rebuilding the loader.
+    private val delegatingCallFactory = Call.Factory { request ->
+        val delegate = currentCallFactory()
+            ?: throw IOException("No server configured; cannot load covers")
+        delegate.newCall(request)
+    }
+
+    val imageLoader: ImageLoader by lazy {
```
Xexanos, line 60: imageLoader is a lazy property first read from MainActivity's setContent (main thread). Building it invokes Coil's diskCache { diskCache } factory, which resolves/creates the cacheDir/covers directory — file I/O on main. Given how carefully this PR isolates the covers dispatcher from playback (RatatoskrClientFactory.kt), it seems worth also keeping this one construction path off main, e.g. by warming coverImages.imageLoader from a background coroutine before/during AppContainer init rather than at first Compose access.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientFactory.kt
```diff
@@ -112,11 +112,21 @@ object RatatoskrClientFactory {
...
+            .build()
+
+        // Each OkHttp stack now owns its own dispatcher thread pool; the shared connection pool
+        // (coversClient reuses mainClient's) is evicted once. Release everything when the client
+        // is replaced so it does not linger until GC (SPEC section 13).
         val closeAction = {
-            for (client in listOf(mainClient, authClient)) {
+            for (client in listOf(mainClient, authClient, coversClient)) {
                 client.dispatcher.executorService.shutdown()
                 client.connectionPool.evictAll()
```
Xexanos, line 134: Minor: coversClient shares mainClient's connection pool by construction (line 121, newBuilder()), so connectionPool.evictAll() here runs redundantly for two of the three clients in the loop. Harmless (idempotent) — just flagging in case it's not intentional.

app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverHeightInterceptor.kt
Xexanos: Fair catch — the comment overstated what the check does. The check itself stays host-agnostic on purpose: knowing the server origin is the wrapper's concern (SPEC §13, contract changes absorbed in one place), and feeding the origin into an app-module interceptor would recreate exactly the coupling the wrapper exists to prevent. If the server ever returns to absolute URLs (the tolerant-reader case), they are its own origin, where appending `?h=` is desired; a truly foreign URL sharing the exact `/library/items/.../cover` path shape would get one extra query parameter, which servers ignore. Comment corrected to say precisely that in 2600b17.

app/src/main/java/io/github/xexanos/ratatoskr/covers/CoverImages.kt
Xexanos: Checked this against the Coil 3.5.0 sources — there is no file I/O on that path. `DiskCache.Builder().build()` only constructs `RealDiskCache` (the one potentially costly branch, the free-space query, runs only for `maxSizePercent`; we set `maxSizeBytes`), and okio's `DiskLruCache` reads/creates its journal lazily in `initialize()`, which is called from `get`/`edit`/`size` — i.e. inside the first fetch, on Coil's background dispatcher. On top of that, `ImageLoader.Builder.diskCache { }` takes a lazy initializer that Coil resolves on first cache access, not in `build()`. So the main-thread construction at `setContent` allocates objects but touches no files; a warm-up coroutine would add lifecycle complexity to avoid I/O that doesn't happen. Resolving.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientFactory.kt
Xexanos: Intentional, and your observation actually goes further than the flagged line: all three clients share one pool (auth and main are built from the same `baseBuilder`, which holds the pool instance), so two of the three `evictAll` runs are idempotent repeats. The loop is kept uniform per client on purpose — if any client ever gets its own pool, the cleanup stays correct without anyone having to remember this asymmetry. The old comment claimed the pool "is evicted once", which was wrong; corrected in 2600b17. Resolving.

PR #99, opened 2026-07-21
refactor(ui): split screens into public stateless Screen + ScreenHost

### LINKED ISSUE #100: Split screens into public stateless Screen + ScreenHost; previews and goldens through public seams only
> *Spec produced from a grilling session on 2026-07-21; all decisions below were confirmed by the maintainer.*

## Problem Statement

Preview and screenshot-golden scaffolding lives inside the production screen files, and the screens' layered structure forces it to cheat to get coverage:

- Extracting previews into sibling files (the PR #99 approach) requires widening production composables from `private` to `internal` — test code dictates production visibility.
- The Now Playing previews cannot reach the real screen layout at all, so they **duplicate** it in a preview-only scaffold. The goldens verify a copy; if the real screen drifts, they stay green.
- The "loading" goldens of Library and Speakers show a **bare screen without the loader**: the 500 ms delayed-visibility gate never elapses in a static render, so the goldens named "loading" do not prove the loading state.
- The cover tile's letterbox goldens render a stand-in `Image` with a hardcoded fit instead of the real cover composable, so they do not guard the component's actual default.

The root cause is shared: the previewable surface is not the public surface, and the delayed-loading gate has no seam.

## Solution

Adopt a repo-wide **Screen / ScreenHost split**. Every screen package exposes a **public, stateless `XScreen(state, …)`** — a pure function of its UI state and the previewable surface — and an **`XScreenHost(viewModel, …)`** that owns ViewModel wiring, navigation effects, snackbars, and polling. All previews move to sibling `XScreenPreviews` files and render **only public seams**; the navigation graph calls the Hosts. A small production hook (a CompositionLocal read by the delayed-visibility gate) lets previews and tests render loading indicators immediately, so every loading golden finally shows the loader — including the cover tile's spinner, driven through the real cover composable via the existing public image-loader CompositionLocal. No production composable needs test-only visibility any more.

## User Stories

1. As a developer, I want each screen's production file to contain only production code, so that reading a screen file is not wading through preview scaffolding.
2. As a developer, I want previews to render the public screen composable, so that no production composable has to widen its visibility for tests.
3. As a developer, I want the stateless screen to be the public API of a screen package, so that the previewable thing and the public thing are the same thing.
4. As a code reviewer, I want screenshot goldens to exercise the real composables, so that a green golden actually certifies the shipped rendering.
5. As a code reviewer, I want the Now Playing preview scaffold's layout duplication removed, so that layout drift in the real screen cannot slip past its goldens.
6. As the UX design owner, I want the "loading" goldens to show the loading indicator, so that the loading treatment (knot loader, cover spinner) is pinned pixel-for-pixel like every other state.
7. As the UX design owner, I want the cover tile's letterbox goldens to go through the real cover composable, so that they guard its actual content-scale default rather than a stand-in's.
8. As a developer, I want a single documented switch that makes delayed loading indicators render immediately in static contexts, so that loading states are previewable without weakening the 500 ms production behavior.
9. As a developer, I want the stateful wrapper to have a distinct name (`ScreenHost`) that does not collide with the `Route` navigation types, so that the vocabulary stays unambiguous.
10. As an accessibility reviewer, I want the a11y check suite to keep rendering every preview — now including loading-indicator states — so that spinner states are accessibility-checked too.
11. As a future contributor (human or agent), I want the split recorded in the SPEC's screen-structure section and its rationale in an ADR, so that the pattern is followed, not re-litigated.
12. As a developer, I want the navigation graph to keep a single wiring point per screen, so that the split does not scatter ViewModel plumbing.
13. As a maintainer, I want goldens for visually-unchanged states to remain byte-identical (git renames) through the refactor, so that the diff proves the split moved code without moving pixels.
14. As a maintainer, I want the whole repo converted in one change, so that no mixed state exists where some screens follow the old pattern and some the new.
15. As an app user, I am unaffected: no visual or behavioral change ships, except that nothing changes — the split is invisible at runtime.

## Implementation Decisions

- **Full split (decision 1):** the stateless layer becomes the public screen composable; a new thin stateful wrapper takes over ViewModel + effects. Not the light variant (keeping the stateful screen public and exposing the content layer beside it), which would create two competing public entry points.
- **Naming (decision 2):** the stateful wrapper is `XScreenHost`. `XRoute` is rejected because `Route` is already the SPEC's name for the `@Serializable` navigation destination types; `Host` matches established Compose intuition (`NavHost`, `SnackbarHost`).
- **Boundary (decision 3):** the public `XScreen` is exactly today's stateless content layer — a pure function of UI state, preview-relevant UI parameters (e.g. the library's query text), and callbacks. The Host owns the ViewModel, query ownership, snackbar presentation, navigation `LaunchedEffect`s, and Now Playing's lifecycle-scoped polling. Snackbars therefore remain non-previewable — accepted; they were never golden-covered, and making them previewable would push Compose infrastructure types into the screen signature.
- **Now Playing (decision 4):** the one real restructure. The public `NowPlayingScreen` takes ownership of the full visual — column, header, and the loading / nothing-playing / loaded branching — which today live in the stateful layer. The preview-only layout duplicate (`NowPlayingPreviewScaffold`) is deleted; previews render the real screen. The loaded-branch content composable stays private inside.
- **Loading-gate seam (decision 5):** the delayed-visibility helper (`rememberDelayedVisible`) reads a new CompositionLocal (working name `LocalImmediateLoading`, final name at implementation time; defaults to off) that, when set, opens the gate immediately. This is the **only new seam** and follows the repo's established idiom of CompositionLocals as test hooks (reduced motion, cover image loader). Production behavior is unchanged when the local is not provided.
- **Cover previews through public seams (decision 5, continued):** the cover tile's loading preview provides a never-resolving image loader through the existing public image-loader CompositionLocal plus the gate local; the letterboxed/loaded previews provide an instantly-succeeding loader with a synthetic bitmap, so the goldens exercise the real cover composable and its real content-scale default. The tile's internal loading-content seam from PR #99 reverts to private.
- **Prototype gate:** whether a never-resolving/instantly-succeeding loader renders the intended cover states under the golden harness's paused clock is unverified. A throwaway prototype answers this **before** the implementation PR. **Fallback if it fails: drop the cover-spinner goldens** (the screen-level loading goldens still work — no async involved). Never an internal-visibility seam.
- **Rollout (decision 6):** all six screens (connect, sign-in, library, speakers, now-playing, settings) in **one PR**, by reworking the open preview-extraction PR (#99) in place — retitle, new body, force-push. No intermediate half-converted state.
- **Documentation (decision 7):** SPEC section 13 gains the Screen/ScreenHost pattern (structure changes go through the SPEC per the repo's working agreement), and the rationale lands as the repo's first ADR. **No** glossary entry: Screen/ScreenHost is code-structure vocabulary, not domain language, and the glossary stays implementation-free.
- The a11y check suite imports preview functions directly; preview names stay stable (or its imports are updated in the same change). Previews remain in the main source set — the preview pane and the golden generator scan only that.
- A `refactor:` change — the version-bump release guard correctly requires no version bump.

## Testing Decisions

- **What makes a good test here:** a golden or check that renders a *public* composable with a fixed UI state and asserts user-visible output. No test reaches into private/internal layers; if a state cannot be reached through a public seam, the seam (not the visibility) is the thing to fix.
- **Screenshot goldens (existing Roborazzi-from-previews pipeline):** every preview continues to generate a golden. For visually-unchanged states the goldens must survive as **byte-identical renames** (only the generated class name changes) — this is the refactor's no-pixels-moved proof, established in PR #99. Intentional pixel changes are limited to: loading goldens that now show the loader, and cover goldens now rendered through the real composable; each gets re-recorded deliberately.
- **New goldens gained:** Now Playing's loading state (previously unreachable), and truthful loading states for Library, Speakers, and the cover tile.
- **Accessibility checks (existing instrumented suite that renders every preview):** automatically extends to the newly-truthful loading states; imports updated with any preview moves.
- **Whole-app flow tests (existing instrumented suite driving the real activity):** untouched by design — the Hosts keep the same wiring the navigation graph always had.
- **Prior art:** the sibling-previews extraction and rename-only golden proof from PR #99; the CompositionLocal test-hook pattern from the cover image loader and reduced motion.
- Full unit suite, golden verification, and the deterministic UX guardrail script must pass; the UX-review skill covers the judgment calls.

## Out of Scope

- Snackbar previews/goldens (stay in the Host, non-previewable for now; revisit in its own decision if wanted).
- Any visual redesign or behavior change — the split must not move pixels except the enumerated loading-golden corrections.
- ViewModel logic, domain models, networking, or anything in the server repo.
- Glossary (CONTEXT.md) changes.
- Changing the 500 ms delayed-visibility production behavior itself.
- Multi-module restructuring; everything stays in the app module.

## Further Notes

- Implementation reworks open PR #99 in place (per maintainer decision) rather than stacking on it; the base is current main (the letterbox fix #97/#98 is already merged).
- The prototype (loading-gate + cover loaders under the golden harness) runs first and its outcome decides the cover-spinner-golden question; its result should be noted on this issue before the PR lands.
- The naming of the gate CompositionLocal is a working title; the implementer picks the final name and documents it in the ADR.

Xexanos:
Implements #100 (spec) — reworked in place from the earlier preview-extraction exemplar, per the maintainer decision recorded there.

## What

Every screen package now exposes two composables:

- **`XScreen(state, …)`** — public and stateless, a pure function of its UI state. This is the surface previews, screenshot goldens, and accessibility checks render.
- **`XScreenHost(viewModel, …)`** — owns ViewModel wiring, navigation effects, snackbars, and polling. Rendered only by the navigation graph. (`Host`, not `Route`: `Route` is SPEC §13's name for the `@Serializable` navigation destination types.)

Previews live in sibling `XScreenPreviews.kt` files (main source set — the preview pane and Roborazzi's generator scan nothing else) and reach **every state through public seams**. The interim `internal` visibility seams (`LibraryContent`, `LoadingTileContent`) are gone; no composable widens its visibility for tests.

## The loading-gate seam (the one new seam)

`rememberDelayedVisible` hides loading indicators for 500 ms — and under a paused composition clock that delay never elapses, so the goldens named "loading" showed **no loader at all**. It now reads `LocalImmediateLoading` (default off; previews/tests provide `true`), the same CompositionLocal-as-seam idiom as `LocalReducedMotion` and `LocalCoverImageLoader`. Production behavior is unchanged.

Newly truthful/new goldens: Library loading, Speakers loading, Connect inspecting, Now Playing loading (didn't exist), and the cover tile's spinner — the latter through the **real `CoverImage`** with a never-resolving loader via the public `LocalCoverImageLoader` (verified deterministic in the #100 prototype).

## Now Playing: the restructure that pays for the pattern

The stateful screen owned the layout (header + loading/empty/loaded branching), so its previews **re-implemented that layout** in a preview-only scaffold — goldens verifying a copy that could not fail when the real screen drifted. The public `NowPlayingScreen` now owns the full visual, the scaffold is deleted, and its three goldens survived as **byte-identical renames**: proof the copy was faithful and the restructure moved no pixels.

## Golden accounting

- All visually-unchanged states: **R100 renames** (only the generated class name changed).
- Content changes: **only** the loading goldens enumerated above (deliberate re-records).
- New: Now Playing loading, Connect inspecting.

## Docs

- SPEC §13 records the pattern; `docs/adr/0001-screen-screenhost-split.md` (the repo's first ADR) records the rationale, the rejected alternatives, and the one irreducible exception: the **loaded**-cover state, which no public seam can golden (falsified twice in the #100 prototype — fake loader and coil's preview handler both never deliver onto the paused-clock frame). Its goldens keep the synthetic stand-in and the ADR says "do not fix this by widening visibility."

## Tests

- `test`, `verifyRoborazziDebug`, UX gate green; androidTest sources compile.
- A11y checks extended to the newly reachable states (loading previews, sign-in submitting, settings unconfigured — the latter two were `private` before and untested; now every preview is a11y-checked, per user story 10).
- Whole-app flow tests untouched (Hosts keep the exact wiring the nav graph always had).
- `refactor:` commits — the version-bump guard correctly requires no bump.

Closes #100


Xexanos:
This PR will be reworked in place to implement the full Screen/ScreenHost concept — spec in #100. The preview-extraction content here survives as part of that change; the `internal` visibility compromises (`LibraryContent`, `LoadingTileContent`) will be resolved by the split.

PR #103, opened 2026-07-23
Docked mini player with shared session state and ticking position

### LINKED ISSUE #79: Library: docked mini player for the active session (screen 03 UX delta)
## What

Implement the docked mini player on the Library screen, as designed in `docs/ux-design.html` screen 03: it exists only while a session is active, shows book, speaker and position, offers pause directly, and opens the Now-playing screen on tap.

## Notes

- The design decision is already recorded in ux-design.html (screen 03, decision 1): the mini player replaces the "Now playing" text/icon button and "feeds off the same `getCurrentSession` polling — no new route, no new module".
- Today the session polling lives exclusively in `NowPlayingViewModel` and only runs while that screen is visible. The mini player needs Library-side session polling (or a shared session-state holder in `AppContainer` that both screens feed from) — that infrastructure is the bulk of this issue.
- **Continue-listening shelf refresh (from #52):** once the Library observes session state, wire the transition *active → none* (playback stopped/finished) to a silent refetch of `GET /library/in-progress`. This closes the one staleness gap left open by #52's re-entry refresh: an externally stopped or naturally finishing session while the user sits on the Library screen. Decided in the #52 grilling session (see the decision record comment there).

## Out of scope

- Media-session / notification / platform integration (#63).

### LINKED ISSUE #101: Spec: docked mini player, shared session state, ticking position (implements #79)
Spec for #79. Decisions were made in the grilling session recorded there (decision record in the #79 comments).

## Problem Statement

While a book is playing, the library screen gives no sign of it. To see what is playing, pause it, or check the position, the user must leave the library through a "Now playing" button — which sits there even when nothing is playing and then leads to an empty screen. And when a session ends while the user sits on the library (someone stops it from the Sonos app, or the book simply finishes), the continue-listening shelf keeps showing the old order until the user happens to leave and come back — the one staleness gap #52 left open.

## Solution

A docked mini player at the bottom of the library, exactly as designed in `docs/ux-design.html` screen 03: it exists only while a session is active, shows cover, book, speaker and position, offers the primary transport action (play/pause toggle) directly, and opens the Now-playing screen on tap. It replaces the "Now playing" header button. The displayed position ticks forward locally while playing instead of jumping in poll steps, and the same ticking derivation drives the Now-playing slider. Underneath, session polling moves out of the Now-playing screen into a shared, process-wide `SessionManager`, and the library uses its session-ended edge to silently refresh the continue-listening shelf.

## User Stories

1. As a listener browsing the library while a book plays, I want a docked mini player showing book, speaker and position, so that the active session stays visible without leaving the library.
2. As a listener, I want the mini player to exist only while a session is active, so that the screen never suggests something is playing when nothing is.
3. As a listener, I want to pause playback directly from the mini player, so that stopping the audio is one tap from the library.
4. As a listener with a paused session, I want the mini player's button to offer resume, so that continuing is just as direct as pausing.
5. As a listener, I want tapping the mini player to open the Now-playing screen, so that the full controls are one tap away from anywhere in the library.
6. As a listener, I want the displayed position to advance smoothly once per second while playing, so that the display doesn't stutter in 5-second poll jumps.
7. As a listener, I want the displayed position frozen while the session is paused or buffering, so that the display never counts time that isn't being played.
8. As a listener on the Now-playing screen, I want the slider to creep forward with the same ticking while I am not dragging it, so that both surfaces tell the same story.
9. As a listener whose book finishes or is stopped externally while I sit on the library, I want the continue-listening shelf to refresh silently, so that the shelf reflects reality without me leaving the screen.
10. As a listener who was searching when the session ended, I want the shelf to be fresh when I clear the search, so that the staleness gap doesn't survive behind a query.
11. As a listener on a flaky network, I want the mini player to keep showing the last known session state during poll failures, so that a blip never masquerades as "nothing playing".
12. As a listener whose pause or resume tap fails, I want a snackbar telling me so, so that a failed command never fails silently.
13. As a listener without an active session, I no longer want a "Now playing" button in the header, so that no control leads to an empty screen.
14. As a listener scrolling a long library, I want the mini player docked (not scrolling away), so that the active session is reachable without scrolling back.
15. As a user who backgrounds the app, I want session polling to stop, so that the app doesn't drain battery or hit the server invisibly.
16. As a user sitting on the speakers or settings screen, I want session state to stay current, so that returning to the library shows a warm, correct mini player immediately.
17. As a user who hasn't connected or signed in yet, I want no session polling at all, so that the app stays silent before it has a server to talk to.
18. As a signed-in user during an active session, I want the client's own token refresh suppressed by exactly one owner of that flag, so that the token-rotation handover (SPEC §5) can't be broken by two screens disagreeing.
19. As a German-speaking user, I want all new mini player strings in German as well as English, so that the app stays fully localized.
20. As a TalkBack user, I want the mini player and its toggle properly described, so that the active session is controllable non-visually.

## Implementation Decisions

- **`SessionManager` (data layer, wired in the manual DI container)** becomes the single owner of session truth: the 5-second poll loop, the transport commands (pause/resume/seek/stop), the command-epoch guard against poll/command races, `NoActiveSession` handling, and the `sessionActive` token-rotation flag. It is the only writer of that flag.
- **Polling is process-wide**, bound to app foreground via `ProcessLifecycleOwner` (new dependency: `androidx.lifecycle:lifecycle-process` — free software, F-Droid-compatible). Foreground polls, background is silent. With no client configured the poll step is a local no-op (no server traffic before trust + sign-in). The poll step itself is externally drivable (same pattern as today's screen-driven `refresh()`); the lifecycle binding is thin wiring above it.
- **ViewModels become thin adapters** over `SessionManager` state. The Now-playing ViewModel keeps the host/screen structure (ADR 0001) and its stopped-navigation trigger; its `onCleared()` flag reset is removed — flag ownership lives in `SessionManager` alone.
- **"Active session" has one definition** — a session in state playing, paused, or buffering (now in `CONTEXT.md`). It gates both the token-rotation flag and the mini player's existence. Stopped/finished tails count as "no active session": the server tears the session down at a natural end itself, so the polled truth converges within one sync tick.
- **Mini player**: exists iff a session is active; shows cover, title, speaker + position line; one play/pause toggle mirroring the Now-playing primary control (playing → pause icon → pause; paused/buffering → play icon → resume); whole surface tap opens Now Playing. Docked at the bottom of the library, per ux-design screen 03.
- **Ticking position is presentation-only**: displayed position = last polled position + elapsed time since it was received, clamped to the duration, only while playing; frozen at the polled value while paused or buffering; re-anchored on every poll. `SessionManager` state remains pure server truth (position + receipt timestamp). Built as one reusable derivation and used by both the mini player and the Now-playing slider (finger-drag still wins; seek on release unchanged). Device-side pauses may tick up to one poll interval too far before snapping back — accepted.
- **Entry point**: the "Now playing" icon button in the library header is removed; the settings gear stays. Without an active session there is deliberately no path from the library to Now Playing (the speaker flow is the other path). Converting the header row to a real M3 `TopAppBar` is out of scope (follow-up issue).
- **The mini player has no error state**: a failed poll keeps the last known session showing, silently — claiming "nothing playing" during an outage would be a lie (same principle as the loading tile). It disappears only on a positive "no active session" answer. A failed pause/resume command surfaces as a library snackbar (no action button — the toggle itself is the retry); the toggle snaps back to server truth on the next poll.
- **Shelf coupling (from #52)**: the *active → not active* edge triggers the existing silent shelf-only refetch. No trigger on cold start (unknown → none). Stop-and-return may double-fetch (edge fires while the library is on the back stack, re-entry refetches again) — accepted, the in-flight-refresh coalescing already bounds it to one visible effect. If the edge fires during an active search, a `shelfStale` flag is set instead and the shelf fetch condition of the full-load pipeline picks it up when the query clears.
- **Documentation shipped with the change**: ADR 0002 recording the process-wide-polling ownership decision (screen-driven polling was weighed and rejected: the flag freezes stale on non-polling screens, and background silence was an accepted limitation, never a requirement); a precision note on ux-design screen 03 decision 5 (toggle, ticking); the "Active session" glossary entry (already in `CONTEXT.md`).
- Poll interval stays 5 seconds, as a `SessionManager` constant. New strings EN + DE.

## Testing Decisions

- Good tests observe external behavior only: state flows, wire requests received by a mock server, and rendered screens — never internals like epoch counters directly.
- **`SessionManager` unit tests at the existing mock-HTTPS-server seam** (real `ConnectionManager`, real generated client, wire fixtures — prior art: the Now-playing and library ViewModel tests). Cover: the active→none transition; the epoch guard (a poll in flight when a command lands cannot revert it); a late poll after stop cannot revive the session; `NoActiveSession` clears state and releases the rotation flag; the flag follows the one active-session definition; no-op without a configured client.
- **ViewModel tests run through the real `SessionManager`** at that same seam — no fake session layer. Library additions: transition triggers a shelf-only refetch (asserted via requests hitting the mock server); no trigger on cold start; `shelfStale` covers the search-time edge; command-failure snackbar state. Now-playing tests adjusted to the adapter shape, semantics preserved.
- **One new seam**: an injected time source for the ticking derivation (a parameter, not a global), so ticking is deterministic under test.
- **Goldens/previews at the existing pure-screen state seam**: library with mini player playing, paused, and absent; unchanged Now-playing goldens guard the refactor.

## Out of Scope

- Media-session / notification / platform integration (#63).
- Converting the library header to an M3 `TopAppBar` component (follow-up issue, own PR).
- Removing `finished` from the contract's `PlaybackState` enum (follow-up issue in the server repo; the app-side cleanup lands after that contract change).
- Ticking anywhere beyond the mini player and the Now-playing slider; list windowing (#65); any change to server polling semantics.

## Further Notes

- The `CONTEXT.md` "Active session" entry is already written (working tree); it should ride with this implementation's PR.
- The decision record for this spec lives in the #79 comments (grilling session, 2026-07-23).

Xexanos:
## Summary

Implements #101 (spec for #79): a docked mini player on the Library screen,
process-wide session polling, and a ticking playback position.

- **`SessionManager`** (new, `data/SessionManager.kt`): the single owner of the
  5s poll loop, transport commands (pause/resume/seek/stop), the command-epoch
  guard, `NoActiveSession` handling, and the `sessionActive` token-rotation
  flag. Bound to `ProcessLifecycleOwner` in `AppContainer` (new
  `lifecycle-process` dependency), so polling follows app foreground/background
  instead of any one screen.
- **`NowPlayingViewModel`** is now a thin adapter over `SessionManager.state`;
  its own poll loop, epoch guard, and `onCleared()` flag reset are gone.
- **`SpeakerManager`** (new): resolves `Session.speakerId` to a display name
  for the mini player, caching the speaker list and refetching only on a
  cache miss. The **Speakers screen** now also loads through it and forces a
  fresh fetch (`refresh()`) on every load - the moment a stale name/membership
  would actually be visible - which keeps the mini player's cache current too
  (previously a renamed speaker would have stayed stale in the mini player
  forever, since a cache hit never refetches on its own).
- **Library mini player**: docked at the bottom of the library, shown iff a
  session is active (cover, title, speaker + ticking position, play/pause
  toggle). Replaces the "Now playing" header button. No error state of its
  own — a poll blip keeps showing the last known session; a failed
  pause/resume from the mini player surfaces as a library snackbar instead.
- **Ticking position** (`ui/common/TickingPosition.kt`): presentation-only
  derivation (polled position + elapsed time since receipt, clamped, frozen
  outside `PLAYING`), with an injected clock as the test seam. Shared by the
  mini player and the Now-playing slider.
- **Shelf staleness fix (#52)**: the active → none session edge silently
  refetches the continue-listening shelf; a `shelfStale` flag covers the edge
  firing mid-search, consumed once the query clears.
- ADR 0002 records the process-wide-polling decision; a precision note was
  added to `ux-design.html` screen 03 decision 5.
- Version bumped 1.7.1 → 1.8.0 (feat-level changes) with a fastlane changelog.

## Test plan

- [x] `SessionManagerTest`, `SpeakerManagerTest`, `TickingPositionTest` (new)
- [x] `NowPlayingViewModelTest`, `LibraryViewModelTest`, `SpeakersViewModelTest`
      updated/extended for the mini player, speaker resolution/refresh, and
      the shelf-staleness edge
- [x] Roborazzi goldens re-recorded (existing Library previews for the header
      change) and added (mini player playing/paused states, with their own
      non-overlapping shelf fixture to avoid a duplicate-speakable-text a11y
      finding)
- [x] `AccessibilityChecksTest` extended with the two new mini player previews
- [x] Full JVM unit suite (`./gradlew testDebugUnitTest`) green, run 6x
      consecutively to rule out flakiness from a ViewModel-teardown leak found
      and fixed during review
- [x] `scripts/check-ux.sh` guardrails pass locally
- [ ] Instrumented suite (`connectedDebugAndroidTest`, incl. `AppFlowTest`) not
      run locally — no emulator in this environment; CI covers it

Closes #101.
Closes #79.


app/src/main/res/values-de/strings.xml
```diff
@@ -41,6 +40,9 @@
     <string name="library_loading_more">Weitere Hörbücher werden geladen</string>
     <string name="library_load_more_failed">Weitere Hörbücher konnten nicht geladen werden. Zum Wiederholen tippen.</string>
     <string name="library_retry">Erneut versuchen</string>
+    <!-- Mini player (issue #101): %1$s Status (Wird wiedergegeben/Pausiert/...), %2$s Lautsprecher, %3$s Position -->
+    <string name="library_miniplayer_subtitle_with_speaker">%1$s auf %2$s · %3$s</string>
+    <string name="library_miniplayer_subtitle">%1$s · %2$s</string>
```
Xexanos: Is it realy necessary to localize this? The string should always be the same regardless of language.

Xexanos: Good catch, and you're right. Split this: `library_miniplayer_subtitle_with_speaker` keeps its translation (it carries a real word, "on"/"auf"), but this string here is just placeholders and a separator with nothing to localize - so I pulled it out of both `strings.xml` files and build it directly in `miniPlayerSubtitle` instead, the same way `formatPlaybackTime`'s "H:MM:SS" separator already lives in code rather than as a resource. Pushed in 89a3bdb.

Issue #117, opened 2026-07-30
Prefactor the auth seam (no behaviour change, contract 1.x)


Xexanos:
## Parent
#114

## What to build
A **pure refactor on the current 1.x contract** that narrows the coupling between the app module and the network module's auth/token internals, so the later contract cutover is a contained, mechanical swap ("make the change easy, then make the easy change"). No behaviour change — the /v1 login + rotation-adoption protocol still works exactly as before, guarded by the existing test suite.

## Acceptance criteria
- [ ] The app module no longer reaches into token-pair internals directly; it talks to the network wrapper through a stable, credential-shaped interface.
- [ ] The token store exposes a credential-shaped accessor (a single opaque credential) alongside the existing pair storage.
- [ ] Token-adoption and the `sessionActive` suppression feed are each isolated behind one clearly-named seam, ready to delete in the cutover.
- [ ] No behaviour change; the full existing test suite passes unchanged on the 1.x contract.

## Blocked by
#116

Issue #116, opened 2026-07-30
Rewrite SPEC §5 to the /v2 Ratatoskr-token model


Xexanos:
## Parent
#114

## What to build
`docs/SPEC.md` §5 currently documents the /v1 login + refresh-token **rotation-adoption protocol** (contract 1.1.0) — the exact protocol #114 removes. Rewrite §5 to the /v2 model per server ADR-0001: the app's only credential is a single **opaque, non-expiring Ratatoskr token**. Doc-only, no code. This is the authoritative spec every other ticket builds against.

## Acceptance criteria
- [ ] §5 describes the single opaque, non-expiring Ratatoskr token as the app's only stored credential (Keystore-backed), alongside server URL, TLS fingerprint, and display username; password never stored.
- [ ] §5 specifies `POST /v2/auth/login` and `POST /v2/auth/logout`, and states there is **no app-side token lifecycle** (no `/auth/refresh`, no rotation adoption, no `sessionActive` suppression).
- [ ] §5 specifies 401 handling: body `code: UPSTREAM_SESSION_LOST` → targeted re-login (server URL, certificate trust, username survive); any other 401 → device signed out.
- [ ] §5 specifies the one-time first-launch /v1→/v2 migration (discard stored ABS tokens, prompt re-login; server/cert/username survive).
- [ ] §5 specifies the 404 /v2 rollout guard ("please update your server").
- [ ] The "Refresh-token rotation (contract 1.1.0)" subsection is deleted; the §4 note on absorbing contract changes in the thin wrapper is reconciled.

## Blocked by
None — can start immediately.

Issue #114, opened 2026-07-28
Migrate to /v2: Ratatoskr token, drop the rotation-adoption protocol


Xexanos:
App-side implementation of the server's [ADR-0001](https://github.com/Xexanos/ratatoskr-server/blob/main/docs/adr/0001-client-auth-ratatoskr-native-sessions.md) (Ratatoskr-native sessions), fully specified in app SPEC §5. **Blocked by the server's contract 2.0.0 tag** (server-side cut issue) — the submodule pin needs that tag to exist.

## Goal

Migrate the app to **`/v2` only**: the app's credential becomes a single opaque, non-expiring **Ratatoskr token**; ABS tokens and the entire rotation-adoption protocol leave the app.

## Requirements (app SPEC §5)

- [ ] Bump the contract submodule to the **2.0.0 tag**, regenerate the client; absorb the change in the thin wrapper (SPEC §4).
- [ ] Sign-in posts ABS credentials to `POST /v2/auth/login`, stores only the returned **Ratatoskr token** (Android Keystore-backed, never plain prefs/logs), plus server URL, TLS fingerprint, display username. No refresh endpoint, no proactive renewal, **no token lifecycle on the app side at all**.
- [ ] **Delete the rotation-adoption protocol and its test matrix** — `rotatedTokens` handling, refresh suppression (`sessionActive`), `/auth/refresh` calls, refresh-token handover at `startSession`. Deleted, not kept dormant (the `/v2` app has no `/v1` fallback).
- [ ] **401 handling (the whole protocol):** body carries `code: "UPSTREAM_SESSION_LOST"` → targeted re-login prompt ("your media-server sign-in expired"); server URL, certificate trust, and username survive. Any other 401 → the device is signed out; return to sign-in.
- [ ] Sign-out calls `POST /v2/auth/logout` (best-effort; an unreachable server still clears locally) and clears the stored token.
- [ ] **One-time migration:** on first start after the update, discard stored ABS tokens and prompt a one-time re-login. Server URL, certificate fingerprint, and username survive.
- [ ] **Rollout guard:** a 404 against `/v2` routes shows a targeted "please update your server" prompt (graceful degradation, SPEC §4).
- [ ] Release guards: versionCode bump + en-US and de-DE Fastlane changelogs (repo convention for `feat`/`fix` releases).

## References

[...]

Issue #118, opened 2026-07-30
Cut the client over to contract 2.0.0 and /v2 sign-in


Xexanos:
## Parent
#114

## What to build
The **atomic cutover** to the /v2 contract (2.0.0 has no /v1, so old and new cannot coexist — this lands as one green step). Bump the contract, wire /v2 sign-in, and delete the rotation-adoption protocol wholesale.

## Acceptance criteria
- [ ] Contract submodule pinned to the **2.0.0** tag; client regenerated; the shape change absorbed in the thin wrapper (no hand-edits to generated code).
- [ ] Sign-in posts ABS credentials to `POST /v2/auth/login` and persists **only** the returned Ratatoskr token (Android Keystore-backed, never plain prefs/logs), plus server URL, TLS fingerprint, and display username; password never stored.
- [ ] The Ratatoskr token is sent as the bearer on every request.
- [ ] The rotation-adoption protocol and its **entire test matrix** are deleted — `rotatedTokens` handling, `sessionActive` suppression, `/auth/refresh` calls, and the refresh-token handover at `startSession` — deleted, not kept dormant (no /v1 fallback).
- [ ] Whole app compiles; the test suite passes against /v2; happy-path sign-in → authenticated request works.

## Blocked by
#117

PR #125, opened 2026-07-31
refactor(auth): narrow the app-to-network auth-token coupling ahead of /v2 (#117)

Xexanos:
## What

A **behaviour-preserving prefactor** (issue #117) that narrows how much of the network module's auth/token internals the app module reaches into, so the later `/v2` Ratatoskr-token cutover (#114) is a contained, mechanical swap — *make the change easy, then make the easy change*.

Stacked on `feat/migrate-v2-ratatoskr-token` (targets it, not `main`), on top of the #116 SPEC §5 rewrite.

## Acceptance criteria

- [x] **AC1** — app module no longer reaches into token-pair internals; it talks to a stable, credential-shaped interface (`CredentialStore`).
- [x] **AC2** — the token store exposes a single opaque `Credential` accessor alongside the existing pair storage.
- [x] **AC3** — token-adoption (`SessionRotationAdoption`) and the `sessionActive` suppression feed (`RefreshSuppression`) are each isolated behind one clearly-named seam, ready to delete at the cutover.
- [x] **AC4** — no behaviour change; the full existing test suite passes **unchanged** on the 1.x contract.

## How

- New opaque `Credential` value type (constructor + value `internal`, so no cross-module caller can read or forge it) and a `CredentialStore` seam (`credential()` / `hasCredential()` / `clear()`). `TokenAccess` extends it and projects the /v1 pair to the single credential.
- App code — `MainActivity`, `SettingsViewModel`, `AppContainer`, `ConnectionManager.requireReauth` — routes through the seam instead of `authSession()` / `tokenStore.clear()`.
- Rotation-adoption moved verbatim into `SessionRotationAdoption`; the refresh-suppression flag+feed into `RefreshSuppression`. Each carries a doc block naming what the /v2 cutover deletes.

## Notes

- `ConnectionManager.tokenStore` stays a public `val`: the pinned `SessionManagerTest` calls `connectionManager.tokenStore.authSession()`, and those tests share the module so `internal` wouldn't hide it. AC4 forces it to stay — the narrowing there is documented convention, not compiler-enforced.
- No test or `FakeTokenAccess` edits (new interface methods have defaults); `git diff` over `*Test*`/`*Fake*` is empty.

## Verification

- ✅ `core-network` + `app` compile (debug)
- ✅ Full JVM unit suite green, unchanged
- ✅ All `androidTest` sources compile against the refactored APIs
- ⚠️ Instrumented tests not run on-device locally (no emulator attached); CI's instrumented jobs cover `FactorySessionRotationComponentTest` etc. The adoption logic moved byte-for-byte, so behaviour is preserved.

Refs #117

### DIFF
--- a/app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
+++ b/app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
@@ -8,6 +8,7 @@ package io.github.xexanos.ratatoskr.data
 import io.github.xexanos.ratatoskr.network.api.RatatoskrClient
 import io.github.xexanos.ratatoskr.network.api.RatatoskrClientFactory
 import io.github.xexanos.ratatoskr.network.persist.ConnectionStore
+import io.github.xexanos.ratatoskr.network.persist.CredentialStore
 import io.github.xexanos.ratatoskr.network.persist.TokenAccess
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.NonCancellable
@@ -17,7 +18,6 @@ import kotlinx.coroutines.flow.asStateFlow
 import kotlinx.coroutines.sync.Mutex
 import kotlinx.coroutines.sync.withLock
 import kotlinx.coroutines.withContext
-import java.util.concurrent.atomic.AtomicBoolean
 
 /**
  * Owns the current [RatatoskrClient], rebuilding it whenever the trusted server or its
@@ -26,11 +26,20 @@ import java.util.concurrent.atomic.AtomicBoolean
  */
 class ConnectionManager(
     val connectionStore: ConnectionStore,
+    // The pair-shaped network handle: passed to the factory (which needs the bearer/refresh
+    // tokens) and used by the tests. App code goes through [credentials], never the pair.
     val tokenStore: TokenAccess,
 ) {
-    // While a playback session is active the server owns refresh-token rotation, so the
-    // client must not refresh independently (SPEC section 5).
-    private val sessionActive = AtomicBoolean(false)
+    /**
+     * The credential-shaped seam app code uses for auth state (SPEC section 5): a presence check
+     * and a clear, nothing pair-shaped. Backed by [tokenStore]; the /v2 cutover reshapes what is
+     * behind this without touching the callers.
+     */
+    val credentials: CredentialStore get() = tokenStore
+
+    // While a playback session is active the server owns refresh-token rotation, so the client
+    // must not refresh independently (SPEC section 5). Isolated in one deletable feed:
+    private val refreshSuppression = RefreshSuppression()
 
     // Guards client construction so concurrent callers (e.g. the poll loop and a screen right
     // after sign-in) cannot each build a client. Two OkHttp stacks would each hold their own
@@ -55,16 +64,16 @@ class ConnectionManager(
     private val _reauthRequired = MutableStateFlow(false)
     val reauthRequired: StateFlow<Boolean> = _reauthRequired.asStateFlow()
 
-    fun setSessionActive(active: Boolean) = sessionActive.set(active)
+    fun setSessionActive(active: Boolean) = refreshSuppression.setSessionActive(active)
 
     /**
-     * Terminal auth failure: stop suppressing refresh, discard the stranded tokens, and signal
-     * the UI to send the user back to sign-in. Idempotent - safe to call from several failing
-     * calls at once. Cleared by [acknowledgeReauth] once the nav host has routed.
+     * Terminal auth failure: stop suppressing refresh, discard the stranded credential, and
+     * signal the UI to send the user back to sign-in. Idempotent - safe to call from several
+     * failing calls at once. Cleared by [acknowledgeReauth] once the nav host has routed.
      */
     suspend fun requireReauth() {
-        sessionActive.set(false)
-        tokenStore.clear()
+        refreshSuppression.setSessionActive(false)
+        credentials.clear()
         _reauthRequired.value = true
     }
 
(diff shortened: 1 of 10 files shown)


Xexanos (commented):
Two-axis review (Standards + Spec) against issue #117. **Verdict: solid prefactor — all four ACs substantially met, no behaviour change found, no hard standards violations.** The findings below are judgement calls; the first is the one I'd act on before merge, the rest are take-it-or-leave-it (some are natural to fold into #118 instead).

**Spec axis:** AC1 met at the call-site level (zero pair-internal references left in `app/src/main`), but see the inline note on the still-public `tokenStore` — the seam is currently advisory, not enforced. AC2, AC3, AC4 met (diff touches zero test files; `hasCredential()`/`suppressed`/`clear()` routings verified behaviour-equivalent; the only visibility change, `AppContainer.tokenStore` → private, has no external references).

**Standards axis:** no documented-standard breaches (no generated code touched, §13 package placement respected, no token logging, no new deps). Remaining findings inline.

app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
```diff
@@ -26,11 +26,20 @@ import java.util.concurrent.atomic.AtomicBoolean
  */
 class ConnectionManager(
     val connectionStore: ConnectionStore,
+    // The pair-shaped network handle: passed to the factory (which needs the bearer/refresh
+    // tokens) and used by the tests. App code goes through [credentials], never the pair.
     val tokenStore: TokenAccess,
```
Xexanos, line 31: **[both axes — the one finding I'd fix before merge]** `tokenStore: TokenAccess` stays public on an app-module class, and the KDoc's "app code goes through [credentials], never the pair" is enforced by convention only — any future app code can still call `connectionManager.tokenStore.refreshToken()`. AC1's narrowing is real at today's call sites but not in the type system, and the public visibility exists solely for `SessionManagerTest` (:303/:324) reaching through it.

Suggestion: make the constructor param non-`val` (or `internal`) and let the tests assert against their own fake / the `credentials` view. That turns the seam from advisory into enforced — and makes the #118 cutover diff strictly smaller.

app/src/main/java/io/github/xexanos/ratatoskr/data/ConnectionManager.kt
```diff
@@ -26,11 +26,20 @@ import java.util.concurrent.atomic.AtomicBoolean
...
 ) {
-    // While a playback session is active the server owns refresh-token rotation, so the
-    // client must not refresh independently (SPEC section 5).
-    private val sessionActive = AtomicBoolean(false)
+    /**
+     * The credential-shaped seam app code uses for auth state (SPEC section 5): a presence check
+     * and a clear, nothing pair-shaped. Backed by [tokenStore]; the /v2 cutover reshapes what is
+     * behind this without touching the callers.
+     */
+    val credentials: CredentialStore get() = tokenStore
```
Xexanos, line 38: **[naming, minor]** The same seam is `credentials` here but `credentialStore` in `AppContainer` (:45). One name would make the seam trivially greppable at cutover time — worth unifying while the diff is this small.

app/src/main/java/io/github/xexanos/ratatoskr/data/RefreshSuppression.kt
```diff
@@ -0,0 +1,31 @@
...
+class RefreshSuppression {
+
+    private val active = AtomicBoolean(false)
+
+    fun setSessionActive(active: Boolean) = this.active.set(active)
+
+    fun isSessionActive(): Boolean = active.get()
+
+    /** The read the client factory hands to the authenticator (a plain background-thread poll). */
+    val suppressed: () -> Boolean = ::isSessionActive
```
Xexanos, line 30: **[naming, minor]** Three vocabularies for one flag: the class says *Suppression*, the accessors say `setSessionActive`/`isSessionActive`, and this property reads like a Boolean state but is a `() -> Boolean` provider. Since the whole class is a to-be-deleted seam (#118), fine to leave — but if you touch it, one vocabulary (e.g. `isSuppressed()`, `asProvider()`) would read cleaner.

Also noting for #118 (no action here): the suppression feed isn't fully one seam — the factory's `sessionActive: () -> Boolean` param and the authenticator wiring live outside this class, so the cutover will still touch both modules for this feed, as the KDoc already concedes.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClient.kt
```diff
@@ -100,19 +104,19 @@ class RatatoskrClient internal constructor(
         execute { libraryApi.getLibraryItem(itemId) }.map { it.toDomain(coverEndpoint) }
 
     suspend fun currentSession(): ApiResult<Session> =
-        execute(sessionEndpoint = true) { playbackApi.getCurrentSession() }
-            .adoptingRotatedTokens().map { it.toDomain(coverEndpoint) }
+        rotation.adopt(execute(sessionEndpoint = true) { playbackApi.getCurrentSession() })
```
Xexanos, line 107: **[dedup, low priority]** `rotation.adopt(execute(sessionEndpoint = true) { … }).map { it.toDomain(coverEndpoint) }` repeats four times (currentSession/pause/resume/seek). A private `sessionCall { … }` helper would shrink both this code and the eventual /v2 delete-diff. Equally fine to skip — #118 deletes the `rotation.` half of the shape anyway.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/persist/TokenAccess.kt
```diff
@@ -6,21 +6,35 @@
...
      */
     fun currentAccessTokenBlocking(): String?
+
+    /**
+     * The credential-shaped projection of the /v1 pair (SPEC section 5): present exactly when a
+     * full session is stored, its value the bearer the client sends. Implementations that store
+     * the pair get this for free; the /v2 cutover replaces the projection with a direct read of
+     * the single Ratatoskr token.
+     */
+    override suspend fun credential(): Credential? = authSession()?.let { Credential(it.accessToken) }
```
Xexanos, line 39: **[test gap, note]** The new default `credential()` projection ships without a dedicated test. That's consistent with AC4 ("suite passes **unchanged**"), so nothing to do in this PR — flagging it so #118 adds coverage when the credential path becomes load-bearing.

Issue #123, opened 2026-07-30
Release guards: versionCode + Fastlane changelogs


Xexanos:
## Parent
#114

## What to build
The release plumbing to ship the /v2 migration, per the repo convention for `feat` releases.

## Acceptance criteria
- [ ] versionCode bumped.
- [ ] en-US and de-DE Fastlane changelog entries added for this release.

## Blocked by
#118, #119, #120, #121, #122

Issue #122, opened 2026-07-30
Rollout guard: 404 on /v2 → "please update your server"


Xexanos:
## Parent
#114

## What to build
Graceful degradation when the app is newer than the server (SPEC §4/§5). A server too old to speak `/v2` answers `/v2` routes with a **bare 404** (no structured contract error body — the guidance is entirely client-side). Show a targeted prompt to update the server, plus a generic downgrade fallback.

## Acceptance criteria
- [ ] A 404 against `/v2` routes surfaces a targeted "your server is too old, please update it (2.0.0 or newer)" prompt rather than a generic error.
- [ ] The prompt also offers the fallback of installing an earlier, `/v1`-capable version of the app — phrased **generically, with no hardcoded version number**.
- [ ] Copy follows repo conventions (en-US + de-DE strings).

## Blocked by
#118

Issue #121, opened 2026-07-30
One-time /v1→/v2 migration on first launch


Xexanos:
## Parent
#114

## What to build
The first start after the update carries the user across from /v1 credentials to a /v2 sign-in without losing their trusted-server setup.

## Acceptance criteria
- [ ] First launch after update detects stored /v1 ABS tokens, discards them, and routes to a one-time re-login prompt ("Ratatoskr was updated…").
- [ ] Server URL, certificate fingerprint, and username are preserved across the migration; the user only re-enters their password.
- [ ] The migration runs once; a subsequent launch with a valid Ratatoskr token goes straight in.

## Blocked by
#118

Xexanos:
Implemented in #130, merged into the `feat/migrate-v2-ratatoskr-token` integration branch (the auto-close keyword only fires against the default branch, so closing manually). All three acceptance criteria are covered by tests: TokenStoreMigrationTest (discard + username survives), ConnectionManagerTest (one-shot notice), and MigrationFlowTest (full pre-filled re-login flow through the UI).

Issue #120, opened 2026-07-30
Sign-out via /v2/auth/logout


Xexanos:
## Parent
#114

## What to build
Sign-out tells the server and always clears locally.

## Acceptance criteria
- [ ] Sign-out calls `POST /v2/auth/logout` (best-effort) and clears the stored Ratatoskr token.
- [ ] An unreachable or failing server still clears the token locally and returns the user to sign-in.
- [ ] The logout request carries the bearer token: `BearerAuthInterceptor`'s blanket `/auth/` exemption is narrowed to `/auth/login` (on contract 2.0.0, logout is the one auth endpoint that **requires** the Authorization header — the server resolves the presented token to the device session it revokes), and a test asserts the `Authorization` header on the logout request. Note: MockWebServer does not enforce auth, so without an explicit header assert a logout test goes green while the server-side revocation silently never happens (see the review discussion on #126).

## Blocked by
#118

PR #126, opened 2026-07-31
feat(auth): cut the client over to contract 2.0.0 and /v2 sign-in (#118)

Xexanos:
Implements #118 (the atomic /v2 cutover, part of #114). Stacks onto the `feat/migrate-v2-ratatoskr-token` feature branch on top of the #117 prefactor — **base is the feature branch, not `main`**.

## What this does
The single, green cutover to the `/v2` contract (2.0.0 has no `/v1`, so old and new cannot coexist):

- **Contract pinned to `v2.0.0`**, client regenerated at build time; the shape change (`AuthTokens`→`AuthSession{token,user}`, dropped `Session.rotatedTokens` / `StartSessionRequest.refreshToken` / `RefreshRequest`) is absorbed in the thin wrapper — no hand-edits to generated code.
- **`/v2` sign-in:** `login` posts Audiobookshelf credentials to `POST /v2/auth/login` and stores **only** the returned opaque Ratatoskr token (Keystore-backed, under a new `ratatoskr_token` key), plus server URL, TLS fingerprint, and display username. The password is never stored; the token is the bearer on every request.
- **Rotation-adoption protocol deleted wholesale** (not dormant): `SessionRotationAdoption`, `RefreshSuppression`, `TokenRefreshAuthenticator`/`TokenRefresher`, the `/auth/refresh` call, the `startSession` refresh-token handover, and `TokenAccess.updateTokens/refreshToken` — along with its entire test matrix (`TokenRefreshAuthenticatorTest`, `FactorySessionRotationComponentTest`). Client base path `/v1/` → `/v2/`.
- The 401 reauth escalation keeps its #108 behaviour, now reading `SessionManager`'s own active-session state instead of the deleted suppression flag.

## Deliberately out of scope (own follow-up issues)
Sign-out still only clears the store locally; `POST /v2/auth/logout` (#120), the `UPSTREAM_SESSION_LOST` targeted re-auth copy (#119), the one-time `/v1`→`/v2` migration prompt (#121), and the 404 rollout guard (#122) are their own issues. **No versionCode bump / changelogs here — that is #123**, which must land before the feature branch merges to `main` or the release guard will fail.

## Verification (all green, locally)
- JVM unit tests, both modules
- core-network component suite (21 tests, real Keystore + TLS pin on emulator)
- whole-app `AppFlowTest` (8 tests — full `/v2` flow: connect → confirm cert → sign in → library → authenticated cover → playback)
- `MinifiedWireSmokeTest` under R8 (3 tests — `/v2` login deserialization survives shrinking)

## Review
Ran the two-axis code review. Spec: all acceptance criteria met, no scope creep. Standards: clean; fixed one stale SPEC §10 claim ("refreshes its access token transparently") that the cutover made false, plus two stale code comments.


app/src/main/java/io/github/xexanos/ratatoskr/data/SessionManager.kt
```diff
@@ -181,23 +177,22 @@ class SessionManager(private val connectionManager: ConnectionManager) : Default
...
-     * refresh path is not suppressed, so an ordinary 401 recovers there instead.
+     * signed in yet - and there a 401 is the ordinary "not signed in" state, not a dead
+     * credential. Escalating an unauthenticated-window 401 would clear the just-stored token and
+     * bounce the app off the Library right after sign-in, so only a 401 raised *while a session
+     * is active* is treated as terminal here. The active state is read from [state], the manager's
+     * own view of playback, so the escalation needs no flag on the connection manager.
      */
     private suspend fun RatatoskrError.maybeRequireReauth() {
-        if (this is RatatoskrError.Unauthorized && connectionManager.isSessionActive()) {
+        if (this is RatatoskrError.Unauthorized && _state.value.active) {
```
Xexanos, line 195: The reauth escalation stays gated on an active session, but `/v2` removes the refresh fallback that used to cover the *unauthenticated-window* 401 — worth an explicit note here about what now happens in that window.

Launch routes to Library on `hasCredential()` alone (`MainActivity`). So a user whose stored token was revoked server-side, or whose upstream ABS session died, while the app was closed lands on Library; the first poll/library call 401s while `state.value.active` is still false (no successful poll yet), so this never escalates. On `/v1` the authenticator's refresh would have recovered exactly this window. Net effect on `/v2`: the user is stranded on an errored Library with no automatic route back to sign-in (only Settings → Sign out).

For a fully-dead credential this dead-end also existed on `/v1`, so it isn't a strict regression — but the single-token model makes it the *only* outcome for a revoked token without playback. Is this the intended interim behaviour pending #119 (`UPSTREAM_SESSION_LOST`) / #122? A one-line comment tying the gate's residual to those issues would make the deferral explicit rather than implicit.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/persist/TokenStore.kt
```diff
@@ -29,49 +31,35 @@ class TokenStore(
...
         val username = prefs[USERNAME] ?: return null
-        return AuthSession(access, refresh, AuthUser(userId, username))
+        return AuthSession(token, AuthUser(userId, username))
     }
 
     override suspend fun save(session: AuthSession) {
         dataStore.edit { prefs ->
-            prefs[ACCESS_TOKEN] = crypto.encrypt(session.accessToken)
-            prefs[REFRESH_TOKEN] = crypto.encrypt(session.refreshToken)
+            prefs[TOKEN] = crypto.encrypt(session.token)
```
Xexanos, line 42: The migration note above (lines 23–25) says a pre-`/v2` install's old pair is "simply absent here", but `save()` only writes `ratatoskr_token`/`user_id`/`username` — it never removes the legacy `access_token`/`refresh_token` entries. After an upgrade + fresh `/v2` sign-in those old encrypted values linger in DataStore indefinitely; only a `clear()` on sign-out wipes them.

Low severity (they're encrypted and useless against a `/v2` server), but given the §11/§14 hygiene posture consider a one-shot orphan-key wipe — either here in `save()` or on first `/v2` launch — or confirm #121 (the migration prompt) is where the cleanup lands.

core-network/src/test/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientTest.kt
```diff
@@ -118,14 +118,17 @@ class RatatoskrClientTest {
...
 
         val result = client.startSession("i1", "s1")
 
         assertTrue(result is ApiResult.Success)
         val body = server.takeRequest().body.readUtf8()
-        assertTrue("request should carry the refresh token, was: $body", body.contains("\"r0\""))
+        assertTrue("request should carry the item id, was: $body", body.contains("\"i1\""))
+        assertTrue("request should carry the speaker id, was: $body", body.contains("\"s1\""))
+        // The /v2 request is item + speaker only; no refresh-token handoff leaks into it.
+        assertFalse("request must not carry a token, was: $body", body.contains("token"))
```
Xexanos: Nit: `body.contains("token")` is a broad substring guard. It works today because the `/v2` request body is `{itemId,speakerId}`, but any future field whose name merely contains "token" would make this false-fail. Since you already assert the item/speaker ids are present, an absence check on the specific dropped field — e.g. `!body.contains("refreshToken")` — would express the intent ("no refresh-token handoff leaks in") without the brittleness.

Xexanos (commented):
## Overview

Clean, single-commit cutover from the `/v1` contract to `/v2`: the token pair (`AuthTokens` → `AuthSession{token,user}`) collapses into one opaque, non-expiring Ratatoskr token, and the whole refresh/rotation apparatus (`TokenRefreshAuthenticator`/`TokenRefresher`, `SessionRotationAdoption`, `RefreshSuppression`, the `/auth/refresh` call, the `startSession` refresh-token handoff, and `TokenAccess.updateTokens/refreshToken`) is deleted wholesale — code and its test matrix. Base path `/v1/` → `/v2/` throughout, and the 401 reauth escalation now reads `SessionManager`'s own `state.value.active` instead of the deleted suppression flag.

**Code quality:** excellent. The change is internally consistent (grepped the whole branch — zero dangling refs to any removed symbol or `/v1/` path), comments and KDoc were updated to match the new reality rather than left stale, and the deletions are genuinely wholesale (no dormant dead code). The reauth-gate refactor from `connectionManager.isSessionActive()` to `state.value.active` is behaviour-preserving: the old flag was written only by `applySession`/stop/`NoActiveSession`, so it tracked `active` exactly. Test coverage stays strong — the real Keystore-backed `TokenStore` round-trip (new `ratatoskr_token` key) is still exercised end-to-end by `AppFlowTest`, and `MinifiedWireSmokeTest` still guards `/v2` login deserialization under R8.

**Conventions/security:** follows the repo rules — generated code untouched (regenerated at build), no tokens logged, contract pinned. The out-of-scope deferrals (#119–#123) are clearly enumerated in the PR body.

Three observations below, none blocking — one behavioural question worth an explicit note, one low-severity hygiene item, one test nit.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/AuthInterceptors.kt
```diff
@@ -28,61 +23,10 @@ class BearerAuthInterceptor(
         if (request.url.encodedPath.contains("/auth/")) {
```
Xexanos, line 23: The rewritten doc comment says the bearer skips "the unauthenticated auth endpoints", but this check exempts **every** `/auth/` path. On contract 2.0.0 the only unauthenticated auth endpoint is `POST /auth/login` — `POST /auth/logout` explicitly **requires** the bearer ("Only a missing or malformed Authorization header is a 401"; the server resolves the presented token to the device session it revokes).

When #120 wires up sign-out, this exclusion will silently strip the Authorization header from the logout call: the request 401s, the client still clears locally (logout is best-effort by design), so the server-side revocation quietly never happens — and no user-visible symptom would ever point here.

Suggest narrowing the exemption now, while it is behavior-neutral (nothing calls logout yet), e.g.:

```kotlin
if (request.url.encodedPath.endsWith("/auth/login")) {
    return chain.proceed(request)
}
```

Alternatively, at minimum leave a pointer to #120 here so the trap is visible when that issue is picked up.

Xexanos (commented):
Reviewed the full cutover diff (39 files). This is a clean, deletion-heavy cutover: the rotation-adoption protocol is removed wholesale rather than left dormant, the shape change is absorbed in the wrapper as promised, and the replacement test coverage (401-surfaces-verbatim, startSession-carries-no-token, reauth-from-own-state) matches what the new model actually needs. The submodule bump lands exactly on the `v2.0.0` / `contract-2.0.0` tag. I verified the `SessionManager` reauth gate's ordering: `_state.value.active` is read before the error is written to state in all three paths (poll/stop/control), and the error path retains the last session, so the gate behaves identically to the deleted flag.

One finding, inline below: the `BearerAuthInterceptor`'s blanket `/auth/` exemption contradicts the 2.0.0 contract's auth matrix and will silently break the server-side half of #120's sign-out when it lands.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/api/AuthInterceptors.kt
Xexanos: Resolution: deliberately **not** fixed in this PR — the code change belongs to #120, whose acceptance criteria now include narrowing this exemption to `/auth/login` and asserting the `Authorization` header on the logout request (MockWebServer does not enforce auth, so only an explicit header assert catches this). Recorded there so the trap is visible where the work happens; resolving.

core-network/src/main/java/io/github/xexanos/ratatoskr/network/persist/TokenStore.kt
Xexanos: The cleanup is part of the migration issue.

core-network/src/test/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientTest.kt
Xexanos: Shouldn't the test parse the json instead of using `body.contains()`?

app/src/main/java/io/github/xexanos/ratatoskr/data/SessionManager.kt
Xexanos: Confirmed — intended interim behaviour. On /v2 the single non-expiring token has no refresh fallback, so a credential revoked or expired while the app was closed, with no active playback, lands on an errored Library with Settings → Sign out as the only manual escape. The automatic recovery (any 401 on a stored credential → sign-out / targeted re-login) is #119, and the too-old-server prompt is #122. Leaving the gate unchanged and deliberately not adding a code comment here — the deferral is already tracked in those issues.

core-network/src/test/java/io/github/xexanos/ratatoskr/network/api/RatatoskrClientTest.kt
Xexanos: Done in ce37d09 — the test now parses the body with Moshi and asserts the exact field set `{itemId, speakerId}`, covering both the expected fields and the absence of any token in one check.

Xexanos: See https://github.com/Xexanos/ratatoskr-app/pull/126#discussion_r3705545427

PR #131, opened 2026-08-08
fix(auth): show 'Wrong username or password' for a rejected login

Xexanos:
## What

A 401 from `POST /v2/auth/login` (wrong credentials) surfaced the shared `error_unauthorized` string — "Sign-in expired. Please sign in again." — which is circular and wrong when the user just mistyped their password on the sign-in screen. The design (docs/ux-design.html, Sign in, decision 4) specifies inline "Wrong username or password" for this case.

With the /v1 → /v2 migration (#121) every upgraded install re-enters a password once, so a mistyped password is about to become a common event.

## How

- New `UiError.WrongCredentials` variant, resolved to a dedicated string (`error_wrong_credentials`, English + German).
- `SignInViewModel.signIn` maps a `RatatoskrError.Unauthorized` failure from the login call to `WrongCredentials`. The mapping deliberately lives in the ViewModel, not the shared `UiError`/`RatatoskrError` mapping: only the sign-in screen knows its 401 rejects just-entered credentials rather than an expired session. The generic "Sign-in expired" copy stays correct for 401s on authenticated calls.
- Non-401 login failures (server error, network, ...) still pass through as before.
- The sign-in error preview now renders the realistic wrong-credentials state; the Roborazzi golden for it is re-recorded in the same PR (accessibility checks also run over the preview).

## Tests

- `SignInViewModelTest`: the rejected-login test now asserts `Error(UiError.WrongCredentials)` exactly, and a new test pins that a 500 from login keeps the `Domain` error.
- `:app:testDebugUnitTest --tests SignInViewModelTest` green locally.


Xexanos:
Superseded by #137: same change, rebased onto feat/migrate-v2-ratatoskr-token (the v2-migration integration branch this belongs to) on a descriptively named branch. This PR targeted main, which does not carry the v2 auth work yet.

PR #146, opened 2026-09-01
feat(ui): brand the error colour roles and unify the banner pattern

### LINKED ISSUE #135: Design system: branded error color roles + document the notice/error banner patterns
## Problem

The app has no branded error colors: `Theme.kt` never sets `error`/`onError`/`errorContainer`/`onErrorContainer`, so every error surface falls through to Compose's Material 3 baseline.

**Corrected after measurement** (the values first filed here were from the Material Theme Builder, not from Compose). Read out of `material3-android 1.4.0`, the ramp Compose actually renders is the older 2021 baseline: light `errorContainer` `#F9DEDC` (Error90), light `error` `#B3261E` (Error40), dark `errorContainer` `#8C1D18` (Error30), dark `error` `#F2B8B5` (Error80), dark `onError` `#601410`, dark `onErrorContainer` `#F9DEDC`.

**And the instrument was wrong.** This issue originally blamed a ~1.02:1 luminance ratio against `surfaceVariant`. WCAG contrast measures luminance only, so it reports ~1.00:1 for two surfaces that differ enormously in colourfulness — which is exactly the dark-mode case. Measured in CIEDE2000 and CIELCh chroma instead, the baseline splits cleanly by theme:

| | ΔE vs the notice card | chroma vs its neighbour |
|---|---|---|
| light `errorContainer` `#F9DEDC` | 4.2 | 10.0 vs 9.4 |
| dark `errorContainer` `#8C1D18` | 19.9 | 55.8 vs 8.5 |

**Dark mode is not broken and needs no change.** Light mode is: the error container is as colourless as the warm neutral beside it, which is the real reason PR #134 had to separate the sign-in notice from the sign-in error on glyphs alone. Light `error` `#B3261E` is a second defect — it sits ΔE 3.6 from the copper primary, so the error accent reads as the primary action.

Two related gaps in `docs/ux-design.html` (the contract for the palette):

1. **The token table is 40% silent.** It lists eight roles per mode; `Theme.kt` wires sixteen. Missing: the four error roles plus `onPrimary`, `onPrimaryContainer`, `onSecondaryContainer`, `surface`, `onSurface`, `onSurfaceVariant`, `outlineVariant`, `surfaceContainer`. Worse, the doc actively disagrees with the code: it quotes `Outline #85736C` in four places, while `Color.kt` has held `#7F6D67` since 0f16a70 darkened it for Compose 1.12's contrast check. Nobody noticed, because a human is the only reader an HTML table has.
2. **The notice/banner pattern is undocumented, and the three banners have not actually converged.** `ConnectScreen` had no glyph at all (fixed by hand in #145), and `ShelfErrorRow` puts `error` on `errorContainer` at 4.1:1 — under the AA floor. The doc's own `shelf-err` mockup uses bespoke colors (`#FFF1EC`/`#F3C7BB`/`#3A211B`) that appear in no token table, plus a border the code has never had.

## Decision

The design session settled on: **an error surface is told apart by how colourful it is, not by which red it is.** Dark already obeys that rule; light gets a new warm ramp inside the same hue family rather than a hue shift (a cool crimson was prototyped and rejected — it separates by leaving the brand instead of using it).

- Light: `error #8C1D18`, `onError #FFFFFF`, `errorContainer #FFC4B6`, `onErrorContainer #4A1108`. `#8C1D18` is Material 3's own Error30, which also serves as dark `errorContainer` — one tonal step for two roles.
- Dark: pinned explicitly to the Compose baseline values above. Same pixels, now owned — an inherited ramp shifts silently on a dependency bump, exactly as `OutlineLight` did.
- One `InlineBanner` replaces all three banners, with metrics carried by the container (`ShelfBand`) rather than the call site, and `BannerKind.ERROR` implying `liveRegion = Polite` — an error appearing after the user acted is currently added to the screen in total silence for TalkBack.
- Doc: two-tier token presentation, a new top-level `Patterns` section, the nesting rule moved there out of the Shape prose, and a drift-guard test that parses the role table so the `#85736C` class of failure cannot recur.
- The per-surface icon-weight grep gate floated below is **rejected**: "one weight per surface" needs to know which icons share a surface, which grep cannot decide. It stays a judgement for the `ux-review` skill, where the doc's iconography rule already lives.

Full measurements and the rejected alternatives: https://claude.ai/code/artifact/089cd771-42ab-4fb3-b579-48eb5873208f

Context: ux-review findings on PR #134.

Xexanos:
Closes #135

## What this changes

The theme never set `error`/`onError`/`errorContainer`/`onErrorContainer`, so every error surface fell through to Compose's Material 3 baseline. Two findings reshaped the fix before it was written, and both are corrected in the issue body:

**The hex values in the issue were the Material Theme Builder's, not Compose's.** Read out of `material3-android 1.4.0`, the ramp actually rendered is the older 2021 baseline — light `errorContainer` `#F9DEDC`, dark `#8C1D18`, dark `error` `#F2B8B5`.

**WCAG contrast was the wrong instrument.** It measures luminance only, so it reports ~1.00:1 for two surfaces that differ enormously in colourfulness. Measured in CIEDE2000 and CIELCh chroma instead, the baseline splits by theme:

| | ΔE vs the notice card | chroma vs its neighbour |
|---|---|---|
| light `errorContainer` `#F9DEDC` | 4.2 | 10.0 vs 9.4 |
| dark `errorContainer` `#8C1D18` | 19.9 | 55.8 vs 8.5 |

So **dark mode was never broken** — a saturated brick against a near-grey — and only light changes. The rule this yields, now written into the design doc: *an error surface is told apart by how colourful it is, not by which red it is.* A cool crimson was prototyped and rejected; it separates by leaving the brand rather than by using it.

## Colour roles

Light gets the brand's own ramp inside the warm hue family: `error #8C1D18`, `onError #FFFFFF`, `errorContainer #FFC4B6` (chroma 23.9, ΔE 11.1 vs the notice card and 6.9 vs `primaryContainer`), `onErrorContainer #4A1108`. Light `error` also moves off `#B3261E`, which sat ΔE 3.6 from the copper primary — the error accent read as the primary action.

Dark keeps today's pixels but is now **pinned explicitly rather than inherited**, because an inherited ramp shifts silently on a dependency bump exactly as `OutlineLight` did for Compose 1.12's contrast check. `#8C1D18` serves as light `error` and dark `errorContainer`: one tonal step for two roles.

## One banner instead of three

The three inline banners had converging intent and diverging code — `ConnectScreen` had no glyph until #145 added one by hand, and `ShelfErrorRow` put `error` on `errorContainer` at 4.1:1, under the AA floor. `InlineBanner` replaces all three (the screens lose 178 lines net).

Metrics come from the container via `LocalBannerPlacement`, which `ShelfBand` owns *together* with the shelf's band tone — so nothing can land in the band with the band's background but a top-level banner's padding, and the three duplicated `.background(surfaceContainer)` modifiers collapse into one place. The retry action stays a parameter, because only the call site knows whether a failure can be retried where it happened.

`BannerKind` is semantic, not a colour switch: `ERROR` carries `liveRegion = Polite`. Today an error appearing after the user taps "Sign in" is added to the screen in complete silence for TalkBack — `liveRegion` appears nowhere in `app/src/main`, and the accessibility framework cannot catch that, because a missing announcement is a behaviour rather than a structural violation.

Both glyphs go outlined, which the doc's own iconography rule already required ("filled for primary/active affordances, outlined elsewhere"). `Icons.Default.WarningAmber` was doubly misleading: `Default` is the *Filled* namespace, but the glyph is the outline triangle, so the import lied about the weight.

Settings' sign-out loses its error tint for copper plus a consequence subtitle, matching the two destructive rows beside it (ux-design: Settings, decision 1). The confirmation dialog the doc promises there is deliberately still not built.

## Design doc

The token table was 40% silent — eight roles listed against sixteen wired — and actively wrong: it quoted `Outline #85736C` in four places while the code held `#7F6D67`. Now two-tier: swatches for the tones a decision was made about, plus a complete role→hex table for all twenty roles per mode.

New top-level **Patterns** section for component rules the token scales don't settle. The nesting rule moves there out of the Shape prose, so it stops living in two places, and "Errors stay local" gets a pointer instead of a copy. The `shelf-err` mockup moves onto the real tokens and loses the border the code never had.

## Verification

`ColorRoleDocsTest` parses the role table and fails when it stops matching `Color.kt`. I checked it can actually fail: reintroducing `#85736C` turns it red. The design doc is declared a test input in `app/build.gradle.kts` — without that the unit tests stay `UP-TO-DATE` when only the doc changes, and the guard would sleep through exactly the drift it exists to catch.

Goldens: six new component goldens for `InlineBanner`, and a `SpeakersErrorPreview` that fills the one error path with no golden at all. **`LibraryShelfLoaded` is unchanged**, which is the proof that `ShelfBand` is a pure refactor.

One prediction of mine was wrong and worth flagging: I expected `SignInNoticeVsErrorDark` to show zero diff as the dark-pin proof. It moved — because the glyph swap touches dark too. I verified the dark palette a different way, by comparing large flat colour regions between the old and new golden: `#1A120F` 221937 → 221937 px, and `#53433D`, `#8C1D18`, `#FFB4A3` all identical. Only glyph pixel counts shifted, because the outlined Info fills less area than the filled one. No dark colour moved.

The icon-weight grep gate floated in the issue is **rejected**: "one weight per surface" needs to know which icons share a surface, which grep cannot decide. Proposing a gate that cannot be written is worse than not proposing one; it stays a judgement for the `ux-review` skill.
## Review findings, addressed

Both review axes independently caught the same top defect: the Patterns section quoted chroma **23.9**, which belongs to the candidate that was *not* chosen — `#FFC4B6` measures **24.4**. It sat in the very paragraph stating the axis to preserve. Worth knowing about the guard: prose is outside its reach, so it checks hex cells and cannot check claims about them.

Also fixed from review: Settings decision 1 promised a confirmation dialog the app does not have and may never get, while the code cited that decision as its authority — it now states the copper-and-subtitle rule and records the dialog as deliberately open. The changelog and `InlineBanner`'s KDoc both claimed more convergence than happened, and are narrowed to what is true. `ShelfBand` moved next to `SHELF_ROW_PADDING` in `LibraryScreen` (the shelf's tone, nesting and metrics were split across two files) and is private there now. `ColorRoleDocsTest` re-parsed `Color.kt` per lookup, and its regex required a trailing comma — a role written without one would have vanished from the guard silently. Plus a dead import, two redundant Gradle default imports, and `assertEquals` in place of a bare `assert`.

Two findings left deliberately open:

- **`#85736C` survives once, at `docs/ux-design.html:386`** — the `logo-knot` symbol's path fill. It is logo artwork that happens to share the old hex, not the `outline` role, so changing it would be pretending it is a token. The three places where the value *did* stand for `outline` are fixed.
- **`SpeakersErrorPreview` goldens an error with no retry**, against the doc's "Retry is always the first offered action". That is a pre-existing defect in `SpeakersScreen`, not one this change introduces; the preview only makes it visible. Filed separately, since full-screen error states were explicitly out of scope here.


app/src/main/java/io/github/xexanos/ratatoskr/ui/library/LibraryScreen.kt
```diff
@@ -531,6 +536,37 @@ fun LibraryScreenHost(
...
+ * together, so nothing can land in the band with the band's background but a top-level banner's
+ * metrics. The shelf is a stack of separate LazyColumn items with no shared parent, so each item
+ * wraps itself - which is exactly why the two have to travel in one composable.
+ *
+ * [contentPadding] is applied inside the background on purpose: the band has to fill the gaps
+ * between its rows, which it cannot do if the caller pads from the outside.
+ */
+@Composable
+private fun ShelfBand(
+    contentPadding: PaddingValues = PaddingValues(),
```
Xexanos: Why don't we use an even spacing on every Band? Is the header that different from the other Items?

Xexanos: 🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

Good catch — horizontally the header is **not** different, and the duplication was real.

The 16 dp side inset is identical on every item; it was just expressed in two places. Headers padded themselves, rows were padded by whatever wrapped them — and that split predates the band: the browse rows below the divider had exactly the same wrapper `Box`. `ShelfBand`'s parameter only made an inherited asymmetry visible.

Vertically the header *is* different, and deliberately: a section label belongs to what follows it, so it sits 6 dp below itself against the rows' 8 dp, with 12 dp of air above to separate it from the section before. That is the one value worth keeping apart, so a single uniform padding for the whole band would flatten a real distinction.

So the inset moved into the item instead, which drops the parameter entirely:

- `LibraryRow` carries `LIST_ITEM_INSET`, so a row looks the same in the shelf band and in the browse list without either caller holding the value.
- The browse wrapper `Box` and `SHELF_ROW_PADDING` are both gone.
- `ShelfBand` is now what it claims to be — a tone plus a placement, nothing else.
- The shelf error banner keeps its inset at the call site, because `InlineBanner` is shared and must not bake in one screen's margins. It takes the same `LIST_ITEM_INSET` a book would, since it stands in a book's slot.

`verifyRoborazziDebug` passes unchanged, so this is pixel-neutral across all 52 goldens.

One thing I did *not* do, because it is a design call rather than a code one: if you would rather every band item including the header share one padding, that is a smaller change than this one — but it moves the label's vertical rhythm and the goldens with it. Say the word if you want it.

Xexanos: 🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

Decided: keeping A — the label keeps its own `top 12 / bottom 6` rhythm, everything else shares `LIST_ITEM_INSET`.

Rendered both variants from real Roborazzi frames to compare (same preview, only `LibrarySectionHeader`'s padding differing between the builds): https://claude.ai/code/artifact/186ebf0a-7706-4ac7-8941-713f59a08cf7

The deciding frame was the browse label rather than the shelf one. Inside the band, the uniform inset only pushed the label against the band's top edge; under the hairline divider, where there is no band tone to hold it, the label ended up sitting directly on the line and read as a caption on the divider rather than as the opening of a section.

No code change from this — the branch already carries A. The parameter is still gone, and the inset is still owned by the item; that part of your question stands regardless of which vertical rhythm the label keeps.

Issue #153, opened 2026-09-02
Iconography: non-transport glyphs are filled, violating the design's one-weight-per-surface rule


Xexanos:
`docs/ux-design.html` (Iconography, "Family / weight") states: *Material Symbols; **filled** for primary/active affordances (transport), outlined elsewhere — one weight per surface.* The app follows the first half and ignores the second: every non-transport glyph is `Icons.Default.*` (= filled).

## Inventory

Transport, correctly filled — `Filled.PlayArrow` / `Filled.Pause` (mini player in `LibraryScreen`, `NowPlayingScreen`), `Filled.Stop` (`NowPlayingScreen`).

Report glyphs, correctly outlined — `Outlined.Info` / `Outlined.Warning` in `InlineBanner`, `Outlined.Warning` in `ErrorState` (`CenteredState.kt`), both pinned explicitly by the Patterns section.

Everything else is filled and should not be: `Default.Person`, `Default.Lock` (×2), `Default.Visibility` / `VisibilityOff` in `SignInScreen`; `Default.Lock` (×2) in `ConnectScreen`; `Default.Settings`, `Default.Search`, `Default.SearchOff`, `AutoMirrored.Filled.MenuBook` in `LibraryScreen`; `Default.Home`, `Default.Refresh`, `Default.Delete`, `AutoMirrored.Filled.ExitToApp` in `SettingsScreen`; `Default.Speaker`, `Default.SpeakerGroup` in `SpeakersScreen`.

## Why this is app-wide, and why "leave it alone" is not on the table

Surfaced while reviewing #152 (#133), which added a filled lock to the sign-in trust chip and a filled eye to the password toggle. That PR deliberately did not switch just those two: the chip's lock sits directly above the password field's pre-existing filled lock, so changing one and not the other would put two weights of the same glyph on one screen.

But the "one weight per surface" half of the rule is **already false today**, and not because of #152. The outlined report glyphs are not confined to some separate surface — they render on the same screens as the filled ones, in ordinary states:

- Sign-in in its error or notice state shows filled `Person`/`Lock`/`Visibility` *and* outlined `Warning`/`Info`. The goldens `SignInErrorPreview` and `SignInNoticeVsErrorLightPreview` show both weights in one frame.
- Library's failure state shows filled `Settings`/`Search` in the header *and* outlined `Warning` (`LibraryErrorPreview`, `LibraryLoadMoreFailedPreview`).
[...]

Issue #151, opened 2026-09-02
Whole-screen error states are never announced to TalkBack


Xexanos:
The inline error banner announces itself when it appears (`liveRegion = Polite` for its error kind, introduced in #146). The whole-screen failure state deliberately sits outside that component — a centred glyph, title, body and retry button rather than a tonal card — and carries no announcement semantics at all. So when a screen swaps from its loader to a full-screen failure, TalkBack gets nothing.

Three places are affected: the library's browse-list failure, the speakers-list failure, and Now Playing's no-session branch, which substitutes the error text into the centred "nothing playing" message and is therefore whole-screen shaped even though its error value is the same one the inline banner shows when a session does exist.

This is split out of #132 because it needs a decision first, where #132's two remaining surfaces are already settled by the banner pattern. The question is what the right mechanism is for a whole-screen content swap:

- **A polite live region on the centred state**, exactly like the banner. Simple and consistent — but the banner's rationale is that "an error is added to the screen after the user acted", and here the error *replaces* the screen, so accessibility focus is likely invalidated rather than merely left behind.
- **A `paneTitle` on the swapped content.** This is what Material3 puts on its own snackbar node alongside the live region, and it is what the platform intends for a region of a window whose content is replaced.
- **Conditioned on how the failure was reached.** A failure on first load arrives before the user acted, so normal traversal reads it; a failure after tapping retry arrives mid-interaction and is the case an announcement is for. That would argue for announcing only the second.

Whatever is decided should land as a line in `CONTEXT.md` beside the existing **Error banner** definition, so the inline / whole-screen distinction carries its announcement rule with it rather than living only in code.

Surfaced while triaging #132.

Issue #132, opened 2026-08-08
Now Playing and load-more failures never adopted the announced error banner


Xexanos:
Most of this issue landed incidentally with #146, which unified the banner pattern: the shared inline banner sets `liveRegion = Polite` for its error kind and nothing for its notice kind, so an error banner announces itself when it appears while a notice — already on screen when the screen opens — stays silent. Sign in, Connect and the continue-listening shelf's failure all render through it and are announced, and a JVM Compose test pins both halves of that behaviour. `CONTEXT.md` now defines **Error banner** with "Announced when it appears". #146 never referenced this issue, which is why it stayed open.

Two inline surfaces never adopted the banner and are still rendered as a bare `Text`, so they carry no live-region semantics and TalkBack stays silent when they appear:

- **Now Playing** reports the failure of a transport command (pause / resume / seek / stop) as a centred line below the transport controls. It arrives mid-session, immediately after the user acted — precisely the case the banner's announcement exists for.
- **The library's browse list** reports a failed "load more" page as a tappable row at the end of the list, whose whole area retries.

Neither surface has a `@Preview`, so neither the screenshot goldens nor `AccessibilityChecksTest` cover them today.

This issue is now scoped to those two. Whole-screen failure states (the library's and Speakers' centred error state, and Now Playing's no-session branch) are a different shape and need a mechanism decision first — tracked in #151. Snackbars need no work: Material3 sets a polite live region, a dismiss action and a pane title on the snackbar node itself.

The original report, written before #146 landed, is in this issue's edit history.

Xexanos:
> 🤖 *This was generated by AI during triage.*

## Agent Brief

**Category:** bug
**Summary:** Report the Now Playing transport-command failure and the library's load-more failure through the shared error banner, so both announce themselves to TalkBack

**Current behavior:**
The shared inline banner component takes a banner-kind enum and decides the announcement itself: its error kind applies `liveRegion = LiveRegionMode.Polite`, its notice kind applies nothing. Sign in's failed login, Connect's failed inspection and the continue-listening shelf's failure all render through it and are announced. Two surfaces still render a bare `Text` instead and therefore carry no live-region semantics, so they are added to the screen in silence:

- Now Playing renders the failure of a transport command (pause / resume / seek / stop) as a centred line below the transport controls, tinted with the error colour role. It appears mid-session, directly after the user acted.
[...]

PR #158, opened 2026-09-02
One chip component for the trust and playback-state pills

### LINKED ISSUE #156: Design system: one chip component for the trust and playback-state pills
🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

Two hand-rolled chips now assemble the same pill independently, and only a code comment keeps them in step.

- `ServerTrustChip` (`ui/auth/SignInScreen.kt`, added in #133/#152): pill `Surface`, `secondaryContainer` / `onSecondaryContainer`, leading 16 dp lock glyph, `Row(horizontal = 12.dp, vertical = 6.dp)` with an 8 dp gap, `labelMedium` at `FontWeight.SemiBold`.
- `StateChip` (`ui/nowplaying/NowPlayingScreen.kt`): the same pill and the same 12/6 + 8 dp metrics, but `surfaceVariant` / `onSurfaceVariant`, a leading 8 dp dot instead of a glyph, `RoundedCornerShape(50)` instead of `CircleShape`, and `labelLarge`.

`docs/ux-design.html`, Patterns, is explicit that a recurring surface is carried by the container and never by the call site ("This is carried in code by the container, never by the call site", with `ShelfBand` as the worked example). That is how `InlineBanner` was built for notices and errors in #135/#146. The chips are the remaining case.

## What to build

- One chip component under `app/src/main/java/io/github/xexanos/ratatoskr/ui/`, owning tone, shape, metrics and label typography. The call site says what fact it is stating and what leads it (glyph or dot); it does not get to decide how the pill looks.
- Both call sites on it, with the two private composables gone.
- A Patterns entry in `docs/ux-design.html` describing the chip the way notices and errors are described there.
- Re-recorded goldens.

## Decision 1: which tone does the playback-state chip take?

P3 in the design says "Ash green marks the secondary - status, chips, success", and the playback-state chip is exactly status. That reads as `StateChip`'s `surfaceVariant` being the drift and the new trust chip having the correct tone. `surfaceVariant` is also already spoken for: Patterns gives it to the **notice** banner, so a status chip on the same tone reads as a muted notice rather than as a chip.

Cost of following the design: the Now-playing chip changes tone, which is a visible design change and re-records its goldens. The chip's leading dot has to move with it - today the muted states tint the dot `onSurfaceVariant`, which is the label's own colour, so on the ash container it becomes `onSecondaryContainer` and keeps that relationship. `PLAYING` / `FINISHED` keep the copper `primary` dot, which is the logo's own copper-on-ash pairing.

**Proposed:** ash green (`secondaryContainer` / `onSecondaryContainer`) for both chips.

## Decision 2: which label role do chips take, and is the typography table or the code wrong?

The typography table's `labelSmall` row assigns "11-12 sp - Semibold, +0.5-2 sp tracking" to eyebrows ("PLAYING ON"), chips and time labels. `labelSmall` appears nowhere in the app. What the code actually does:

| Use | Role | Reality |
| --- | --- | --- |
| Eyebrow "PLAYING ON" | `labelMedium` | 12 sp, `letterSpacing = 2.sp` |
| Trust chip | `labelMedium` + SemiBold | 12 sp |
| State chip | `labelLarge` | 14 sp |
| Time labels | `labelLarge` monospace | 14 sp |

So the table's *role name* is wrong and its *size range* is right: 11-12 sp is `labelMedium`, not `labelLarge`. The state chip at 14 sp is the outlier.

**Proposed:** chips take `labelMedium` at SemiBold - the trust chip's value, inside the range the table already asked for - so the Now-playing chip's label drops 14 sp to 12 sp. Time labels stay `labelLarge` and move out of that row: they are 14 sp monospace, and shrinking a ticking clock read at arm's length would cost more than the consistency is worth.

The table is wrong more broadly than this row, though. It documents four roles, one of them unused; the app references nine (`bodyMedium`, `bodySmall`, `labelMedium`, `titleMedium`, `labelLarge`, `headlineSmall`, `headlineMedium`, `titleSmall`, `bodyLarge`). Whether `headlineMedium`, `titleSmall` and `bodySmall` are sanctioned or are themselves drift is a separate consolidation question, not one this issue settles - but the table should at least state what the app does rather than something no longer true of it.

## Guard

`ColorRoleDocsTest` parses the doc's colour-role table and fails when it disagrees with the code, because "a human is the only reader an HTML table has". The typography table is the same kind of claim and had gone the same way silently. Add a sibling guard that reads the typography table and asserts, in both directions, that every role it names is referenced under `ui/` and that the sp it quotes is the sp that role actually carries in `theme/Type.kt`. That is what would have caught the `labelSmall` ghost.

## Done when

- One chip component owns tone, shape, metrics and label typography; both call sites use it.
- `docs/ux-design.html` has a Patterns entry for it, and the typography table no longer contradicts the code.
- The typography drift guard is in place and green.
- Goldens re-recorded; `scripts/check-ux.sh`, `lintDebug` and `verifyRoborazziDebug` green.

Xexanos:
Closes #156.

The two hand-rolled chips - sign-in's `ServerTrustChip` and now-playing's `StateChip` - shared the 12/6 dp padding and the 8 dp gap and had already drifted apart on tone, radius and label size, with a code comment as the only thing holding them together. `InlineBanner` is the precedent: Patterns says a recurring surface is carried by the container and never by the call site, and the chips were the remaining case.

## What is here

- `ui/StatusChip.kt` owns the tone, the shape, the metrics and the label role. A call site passes the copy and says whether a glyph or a dot leads it (`ChipLeading.Glyph` / `ChipLeading.Dot`); it cannot decide how the pill looks. The dot's two tones are a `ChipDot` **emphasis**, not a colour, so no call site can tint one off the palette - and `MUTED` reads the surface's own content colour, so the dot and the label cannot drift apart.
- Both call sites on it; the two private composables are gone. Now-playing keeps the state-to-copy mapping, which is the only part that is screen knowledge.
- Patterns gains a **Chips** entry written the way notices and errors are, plus a delta note in the Now-playing section.
- The typography table is rewritten and `TypographyDocsTest` now holds it to the code.
- Goldens: four new `StatusChip` previews (both themes), and the two Now-playing goldens re-recorded. The sign-in goldens are byte-identical, which is the check that the trust chip's own rendering did not move.

## The two decisions, and what settled them

**Tone: ash green.** The design turned out to be less ambiguous than the code. P3 gives status to the secondary, and the now-playing **mock's own CSS** renders the state chip as `.chip.chip-ash` - `#D1E8B0` on `#121F04`, i.e. `secondaryContainer` / `onSecondaryContainer`. `surfaceVariant` was drift, and it is also spoken for: Patterns gives it to the notice banner, so a status chip on that tone reads as a muted notice. This is the visible change in the PR and the reason the Now-playing goldens re-record.

**Label: `labelMedium` at SemiBold.** The mock's `.chip` rule is `font-size: 12.5px; font-weight: 600; gap: 7px; padding: 0 12px`, and its glyphs are 14 px - the trust chip's values almost exactly. So the table's `11-12 sp` was right and its role *name* was wrong: that range is `labelMedium`, not the `labelSmall` the table named and the app never uses. The state chip's `labelLarge` was the outlier and drops 14 sp to 12 sp. Time labels stay `labelLarge` and move out of that row - they are 14 sp monospace, and shrinking a clock read from across the room costs more than the consistency is worth.

## The drift guard

Yes, worth having, and it works the way `ColorRoleDocsTest` does. `TypographyDocsTest` reads the `#typography` table and asserts, in both directions, that every role it names is one the app really reaches for and that the sp it quotes is the sp that role really carries. Weight is deliberately out of scope: it is a `fontWeight` on the `Text`, not a theme value, so there is no single value to compare against.

The table needed more than a new chip row to pass it. It documented four roles, one of them a ghost, where the app reaches for nine. It now lists all nine with their real sizes.

## Two things I did not do

- **Whether all nine roles are sanctioned** is a separate question. `headlineMedium`, `titleSmall` and `bodySmall` may themselves be drift from a deliberately tight scale; the table now states what the app does, which is the starting point for asking that, not an answer to it.
- **The mock's pulsing dot** (Now-playing decision 5) is still unbuilt, and that is what forced the one place where I did not follow the mock: its dot is ash `secondary` (`#B5D18B`), which against `onSecondaryContainer` would leave `ACCENT` and `MUTED` nearly indistinguishable in dark. The copper `primary` the code already used separates them by hue in both themes, and the label text carries the state regardless, so colour is never the only carrier. Once the pulse exists, the dot's tone is worth revisiting.

Also noticed in passing and left alone: `fastlane/.../2-nowplaying.png` is stale for reasons predating this PR - it still shows the initials cover placeholder that #78 replaced with the knot mark. Recapturing it needs an emulator and would fold an unrelated visual update into this branch.

## Checks

`scripts/check-ux.sh`, `check-version-bump.py` (patch: 2.2.2 -> 2.2.3, changelog pair present), `check-store-metadata.py`, `lintDebug` and `verifyRoborazziDebug` all green locally.


Xexanos:
🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

Ran the repo's `ux-review` agent over the diff. It confirmed both headline moves as corrections rather than drift (it measured `.chip-ash` against `theme/Color.kt` byte for byte, and the label contrast actually *improves* in dark, 5.51:1 → 6.78:1), and found three real problems in the prose this branch added. Fixed in 5e8e418.

**1. Patterns contradicted the document's own rule.** I had written "copper is the one tone that keeps them apart by hue", two paragraphs below "colour must never carry meaning by itself regardless". Measured, the two dots are 2.73:1 apart in light and **1.29:1** in dark - same lightness, separated by hue only, and red/green is the pair colour-vision deficiency takes first. The mock's ash `secondary` is no better (1.27:1), so there was no better tone to pick; what was wrong was the claim. The carrier was never the dot - the label names the state in words, which is also all TalkBack reads. That is now what Patterns, the now-playing delta and the component's KDoc all say, and copper is presented as the compromise it is, including that it is a third copper mark on a screen that already has the play/pause control and the slider.

**2. The weight column was wrong, in the column I had deliberately left unguarded.** Material 3 ships `titleMedium` and `titleSmall` at **Medium**, not Regular, and no call site overrides them. Corrected - and the weight column is now *inside* `TypographyDocsTest`, because guarding the sp column while leaving weight beside it is precisely how that error got in. Call-site overrides moved into the prose column, which stays outside the guard. The guard passes with title/label at Medium and headline/body at Regular, so that is now pinned rather than asserted.

**3. Button labels had fallen out of the table.** The old table mentioned buttons (under `titleMedium`, which was itself wrong); my rewrite dropped the mention without adding them to `labelLarge`. They are back, flagged as applied by M3 rather than named in code - which is also why neither the table nor the guard can see them, so the heading is now "every role the app **names**".

Smaller ones, same commit: the "a chip is never interactive" definition collided with the now-playing Stop control and the "never neither" invariant with the designed Speakers "Last used" chip, so both are named where the definition is made; the leading row now says 16/8 dp round the mock's 14/7 px onto the 4/8-point grid; and the English changelog said "at the same size" of a label that shrank 14 sp → 12 sp.

Also added the one dark now-playing golden the review asked for - that is the screen this change lands on and the theme where the dot separation is weakest.

**Not acted on, deliberately:**

- **No `liveRegion` on the chip**, so a state change is silent to TalkBack. The review raised it as a question, not a finding, and it is right to: nothing in the design or SPEC §12-13 asks for it, it is pre-existing, and it overlaps #151. Worth its own issue rather than this branch.
- **BUFFERING / FINISHED / STOPPED / UNKNOWN are unpinned by goldens.** Pre-existing; `nowplaying_state_unknown` rendering as a dot plus a dash in a pill is worth a look, but it is a state question, not a chip question.
- **Whether all nine typography roles are sanctioned.** Unchanged from the PR description: the table states what the app does, which is where that conversation starts.

`lintDebug`, `verifyRoborazziDebug`, `check-ux.sh`, `check-version-bump.py` and `check-store-metadata.py` green after the round.

Xexanos:
🤖 Heads-up from the session that merged #157 (icon weight, #153) — this PR will go red on rebase, for two independent reasons.

**The new UX gate rule will reject the chip's lock.** `scripts/check-ux.sh` now has a rule 6 that fails on any filled glyph outside the transport set (`PlayArrow|Pause|Stop|Replay10|Forward10|SkipPrevious|SkipNext`). It matches per occurrence, so the three added lines will each be flagged: the `import androidx.compose.material.icons.filled.Lock` and both `ChipLeading.Glyph(Icons.Default.Lock)` call sites. Fix is a one-liner — switch to `androidx.compose.material.icons.outlined.Lock` / `Icons.Outlined.Lock`. Sign-in's own trust chip already moved to the outlined lock on main, so the shared chip should match it anyway.

**Goldens.** #157 re-recorded 26 goldens across every screen that draws an icon. None of the 7 this PR touches overlap, so no binary conflict — but rebase onto main before re-recording, or the PR stays red.

GitHub already reports this branch as `CONFLICTING` against main independently of the above.

Xexanos:
🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

Rebased onto `main` (#157, the icon-weight release). Three collisions, all resolved; force-pushed as `a02f502`.

**The trust chip's lock is now outlined.** #157 changed `Icons.Default.Lock` to `Icons.Outlined.Lock` *inside* the `ServerTrustChip` this branch deletes, so git could not merge it. I kept the deletion and carried the intent to the call site, which is where the glyph now lives. Worth flagging that this was not a free choice: main's new UX-gate rule 6 fails on any filled glyph outside the transport set, so the merge as git would have left it (`Icons.Default.Lock` at the call site) would not have compiled *or* passed the gate. Rule 6 is green. Same fix in the preview file, which is why two component goldens re-record (`5c50bbc`).

Pleasing detail: main's own 20201 changelog already promises "the lock beside the server name … is drawn as an outline", so the resolution matches shipped copy rather than just compiling.

**The release moved to 2.2.2.** 20201 went to the icon-weight release while this branch was open, so `versionCode` is now 20202 and my changelog pair lives at `20202.txt`; main's 20201 pair is untouched. `check-version-bump.py` reads 2.2.1 -> 2.2.2, changelog present. The PR description's version line is corrected.

**`docs/ux-design.html` merged cleanly** - #157 touched the iconography footnote, this branch the typography table and Patterns. I checked main's line survived intact rather than trusting the auto-merge.

**No golden conflicts at all**, which is the part worth noticing: main re-recorded 24 goldens across sign-in, library, settings, speakers and connect; this branch had only touched now-playing. And after re-recording on the rebase, **the sign-in goldens stay byte-identical to main's** - so the extracted chip still renders exactly what the private one did, now with the outlined lock.

`lintDebug`, `verifyRoborazziDebug`, `assembleDebug`, `check-ux.sh` (all six rules), `check-version-bump.py` and `check-store-metadata.py` green on the rebased head.

Xexanos:
🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

Merged `main` in (`7d1e850`) - it had taken #159 (announced error banners, issue #132) and #162. Four conflicts, all resolved by keeping both sides; no force-push, so the review history above stays addressable.

**The two changes actually meet in one frame.** #159 added a `NowPlayingCommandFailedPreview`; that screen now also carries the ash-green chip, so its golden re-records - the only golden this merge touches. Worth a look, because it is the case the Patterns entry argues about: the `errorContainer` banner and the `secondaryContainer` chip sit in the same frame and read as plainly different surfaces. Had the chip stayed on `surfaceVariant`, it would have been the notice tone in a frame with an error report.

**`NowPlayingScreen`'s imports** are the union of main's `BannerKind`/`InlineBanner` and this branch's `ChipDot`/`ChipLeading`/`StatusChip`. One subtlety git could not see: main also moved `common.CoverImage` into its alphabetical slot, so this branch's copy of that import is dropped rather than left as a duplicate.

**Both new previews are kept** - main's command-failed one in state order, this branch's dark showcase after it. The shared `NowPlayingPreview` helper keeps the `dark` parameter this branch gave it, which main's preview neither needs nor minds.

**The release moves to 2.2.3.** Second time this branch has stepped aside: 20201 went to the icon-weight release, 20202 to this one. Main's 20202 pair is untouched, mine is now `20203.txt`, and the guard reads 2.2.2 -> 2.2.3. The PR description's version line is corrected again.

Also checked rather than assumed: main's new `NowPlayingScreenTest` covers the error banner only, so the chip change does not touch its assertions.

`lintDebug`, `verifyRoborazziDebug`, `assembleDebug`, `check-ux.sh` (six rules), `check-version-bump.py` and `check-store-metadata.py` green on the merge.

One note on the watcher: it also fired for a conflict on `a02f502` while GitHub was still recomputing mergeability after the earlier rebase. That one was a stale read - `origin/main` was an ancestor of the head and GitHub's own live answer was `MERGEABLE` - so I verified instead of committing an empty merge. This one was real.

PR #159, opened 2026-09-02
fix(a11y): announce the transport and load-more failures

### LINKED ISSUE #132: Now Playing and load-more failures never adopted the announced error banner

Xexanos:
🤖 *This PR was written by an AI agent.*

Closes #132.

The two inline surfaces that never adopted the shared banner in #146 were still a bare `Text`, so they carried no live-region semantics and TalkBack stayed silent when they appeared. Both now report through `InlineBanner` with its `ERROR` kind, which owns the announcement decision, the warning glyph and the tonal error container.

**Now Playing** reports a failed transport command as a plain error banner below the controls. No action: re-issuing the command *is* the retry, so a retry affordance would only duplicate the control the user just pressed.

**The library's load-more footer** keeps its tap-anywhere retry by passing a `BannerAction` rather than wrapping the banner in a separate clickable, so it stays a single tap target with the label as a second line — the same shape the shelf failure already had. Its message drops the "Tap to retry." the action label now carries.

Both sit in a `TopLevel` container (the footer is outside `ShelfBand`), so neither call site names its own metrics.

## The retry label is now shared

A second actionable banner is what turned the per-screen retry label into a duplicated one: `library_shelf_error_retry` and a new `library_load_more_retry` would have been byte-identical in both locales. It is now one `action_tap_to_retry` in the shared block, alongside `action_retry`, used by both banner call sites and the component's own previews. No rendered text changes, so no golden moved for it.

## The lazy-list question

The brief asked to confirm the load-more banner announces on the failure rather than on being scrolled to, and to report rather than engineer around it if the latter turns out to be true. **It cannot be distinguished at this layer, and the answer is reported on the issue** ([comment](https://github.com/Xexanos/ratatoskr-app/issues/132#issuecomment-5509326801)). In short: `LiveRegion` is consumed in exactly one place in Compose — it sets `info.liveRegion` on the `AccessibilityNodeInfo` — so both "the state flipped" and "the item scrolled in" reach the accessibility service as the same thing: a node with a polite live region entering the tree. This is a pre-existing property of the banner pattern inside a `LazyColumn` (the shelf-failure banner merged in #146 has it too), not something this change introduces. Suppressing the scroll-in case would need screen-level "already announced" state — the separate decision the brief put out of scope.

## Tests

- `NowPlayingScreenTest` and `LibraryScreenTest` (new, JVM Compose — Robolectric, junit4 v2 rule, a semantics matcher on the live-region key, following `InlineBannerTest`): each asserts exactly one polite live region on the failure state and that it carries the failure's own text, plus a negative case (no failure, and the still-loading footer, announce nothing). `LibraryScreenTest` also asserts the banner itself is the retry target.
- `NowPlayingCommandFailedPreview` added with its golden and its `AccessibilityChecksTest` entry; the load-more preview and check already existed, its golden re-recorded. `AccessibilityChecksTest`: 30/30 on the emulator.
- `testDebugUnitTest`, `verifyRoborazziDebug`, `lintDebug`, `scripts/check-ux.sh`, `check-store-metadata.py` and `check-version-bump.py` all pass locally.

Patch bump to 2.2.2 (20202) with both changelogs, as the release guard requires for a `fix`. (#157 released 2.2.1 while this branch was open, so `main` is merged in and the entry moved up a version; the load-more golden was re-recorded after the merge and now carries main's outlined search and settings glyphs alongside the banner.)

## Noticed, not fixed

`docs/SPEC.md` §9 and `docs/testing.md` still describe the JVM level as mappers and ViewModel transitions; `SignInScreenTest` and `InlineBannerTest` already sat outside that description before this PR, and these two add to it. A one-sentence doc correction, but pre-existing and outside #132.


Xexanos:
🤖 Heads-up from the session that merged #157 (icon weight, #153): one of this PR's two goldens conflicts with that merge.

#157 re-recorded 26 goldens across every screen that draws an icon, and `LibraryScreenPreviewsKt.LibraryLoadMoreFailedPreview.Library_-_load_more_failed_W360dp_H800dp.png` is in both sets. Git cannot auto-merge a PNG, so rebase onto main, take main's copy, and re-record:

```
git fetch origin main && git rebase origin/main
git checkout --ours "app/src/test/screenshots/io.github.xexanos.ratatoskr.ui.library.LibraryScreenPreviewsKt.LibraryLoadMoreFailedPreview.Library_-_load_more_failed_W360dp_H800dp.png"
```

No icon changes needed here — this PR adds no filled glyphs, so the new UX gate rule 6 (filled reserved for transport) passes as-is.

Issue #133, opened 2026-08-08
Implement the remaining Sign-in design-card decisions (trust chip, helper, password toggle)


Xexanos:
PR #131 implemented decision 4 of the ux-design "Sign in" card (inline "Wrong username or password"). The other three decisions on the same card (docs/ux-design.html, section 02) are still unimplemented and untracked:

1. **Server trust chip** at the top (lock icon + host + "trusted") linking the credentials to the certificate decision from the Connect step.
2. **Helper text** "You'll stay signed in on this device." below the password field.
3. **Password visibility toggle** (eye icon) — long generated passwords are common with Audiobookshelf, and typing them blind produces failed attempts. This is the decision that *prevents* the wrong-credentials error 4 reports, so it is arguably the more valuable half.

Also called out in the design doc's own delta list for this screen, pre-existing drift:

- `signin_subtitle` reads "Use your existing username and password." where the design says "Use your Audiobookshelf username and password."
- The Sign in button is not pinned to the bottom thumb zone.

Surfaced by the ux-review pass over PR #131.

Xexanos:
> *This was generated by AI during triage.*

## Agent Brief

**Category:** enhancement
**Summary:** Implement the three unbuilt decisions on the Sign-in design card (trust chip, stay-signed-in helper, password visibility toggle) and clear the two deltas the same card records (subtitle copy, bottom-pinned button).

**Current behavior:**
The sign-in screen is a single vertically scrolling column: logo, title, subtitle, an optional notice card, username field, password field, an inline error card, and the Sign in button as the last element in the scroll region.

- There is no server context anywhere on the screen. Nothing links the credentials the user is about to type to the certificate they trusted one step earlier. Repo-wide there is no chip component other than the playback `StateChip`, so this is new.
- There is no helper text below the password field.
- The password field applies `PasswordVisualTransformation` unconditionally and has no trailing icon, so a long generated Audiobookshelf password can only be typed blind.
- `signin_subtitle` reads "Use your existing username and password." The design says "Use your Audiobookshelf username and password."
- The Sign in button scrolls with the form instead of sitting in the bottom thumb zone. No screen in the app currently pins a bottom action, so there is no in-repo pattern to copy.

Decision 4 of the same card (the inline "Wrong username or password" error) is already implemented and must keep working.

**Desired behavior:**

[...]

PR #160, opened 2026-09-02
fix(connect): pin the action to the thumb zone, whichever the state offers

### LINKED ISSUE #155: Connect: the primary action never moved into the bottom thumb zone
#133 moved the sign-in screen's primary action out of the scroll region into the bottom thumb zone, making it the app's first bottom-pinned action. Connect — the screen immediately before it in the first-run flow — still renders every action inline at the end of its scrolling column (`ConnectScreen.kt`). So the inconsistency sits across two consecutive screens of the first run, and it is Connect that lags, not sign-in that deviates: Connect's own design card already asks for the pinned slot. The idle mock's caption reads "Idle — the logo as sender, the action in the thumb zone", and in *both* Connect mocks the actions sit in a `flex:none` block at the bottom of the body, below a `flex:1` spacer.

Pinning also fixes a live defect. At 360×800 the `Confirm` state already overflows: the checked-in golden `Connect - confirm certificate` shows "Trust and continue" jammed against the bottom edge and "Cancel" entirely below the fold. On the one screen whose design card says "this is the one decision here" (decision 3), the decision is barely reachable and its escape hatch is invisible until the user scrolls.

## The decision this needs first

Connect's button is one of five state branches (`Idle`, `Inspecting`, `Confirm`, `Trusted`, `Error`), and the `Confirm` branch carries its own "Trust and continue" plus "Cancel" *inside* the certificate card. So "pin the button" is ambiguous in a way it was not on sign-in, which has exactly one action.

**Decision: the pinned slot belongs to the screen, not to the idle state — it holds whichever action the current state offers, and the certificate card's two actions move into it.**

| State | Pinned slot |
| --- | --- |
| `Idle` | Connect |
| `Inspecting` | *empty* — nothing to confirm yet; the loader is the whole content |
| `Confirm` | Trust and continue + Cancel, lifted out of the certificate card |
| `Trusted` | Connect (unchanged: transient, the navigation effect leaves it immediately) |
| `Error` | Try again |

Three reasons:

- **It is what the design shows.** The Confirm mock puts "Trust and continue" and "Cancel" in the same bottom block the idle mock puts "Connect" in, outside the certificate card, with a `flex:1` spacer above. Pinning only the idle action would build half of one mock and none of the other.
- **Pinning only `Idle` would make the drift worse, not better.** The primary action would jump from the thumb zone (idle) into the middle of a card (confirm) and back down to the end of the scroller (error) — three positions within one screen, where today there is at least one consistent (wrong) position.
- **It is what fixes the overflow.** The `Confirm` overflow above is not a copy-length accident; it is what happens when a fixed-height card ends in two buttons at the bottom of a scroller. Only lifting them out puts them on screen at every height.

The certificate card is then facts only — subject, issuer, validity, fingerprint, the compare hint — which is what decision 1 ("the fingerprint is the centerpiece") wants of it anyway.

The error state stays consistent with the "retry follows the failure's scope" rule: a flow-blocking failure keeps its own full-width button rather than a tappable banner. The rule is about the *scope* of the retry, not about the button's adjacency to the banner, and the button is still below the banner once pinned.

## Out of scope

Three further gaps on the same design card are deliberately **not** in this issue; they belong in the card's Δ list rather than in a layout change:

- The idle mock also pins the **URL field** and a "Host and port, HTTPS only." helper inside the same bottom block, centring logo, title and subtitle in the space above. Keeping the field in the scroll region matches sign-in (whose fields scroll under its pinned action) and Compose brings a focused field into view by itself; the helper line is a separate copy decision.
- **Decision 4, "the logo appears only in the idle state", is still unbuilt** — the logo and the URL field render in every state, including `Confirm`, where the mock replaces them with a top app bar and an intro line naming the host. That is a layout question of its own, though the pinned slot is a prerequisite for it.
- The fingerprint is still one long wrapped string, not the mock's four groups of eight.

## Done when

- [ ] Connect's actions live in a bottom-pinned slot outside the scroll region, following the `SignInScreen` pattern: outer `Column(Modifier.fillMaxSize().imePadding())`, scrolling content in a child with `Modifier.weight(1f).verticalScroll(...)`, the action after it with the design's 32 dp group break (8 dp inside the scroller plus 24 dp on the action).
- [ ] The certificate card holds facts only; both of its actions are in the pinned slot.
- [ ] A preview covers the scroll/action boundary at a height where the content genuinely scrolls under the pinned action (sign-in's `Sign in - scrolled` is the model), and it runs in `AccessibilityChecksTest`.
- [ ] Connect goldens re-recorded; `AccessibilityChecksTest` green in both light and dark.
- [ ] The Connect design card's Δ list records what this builds and what stays open.

Xexanos:
#133 pinned sign-in's primary action to the bottom thumb zone, making it the app's first bottom-pinned action. Connect — the screen immediately before it in the first run — still rendered every action inline at the end of its scrolling column, so the inconsistency sat across two consecutive screens. It was Connect that lagged, not sign-in that deviated: Connect's own design card asks for the pinned slot ("Idle — the logo as sender, the action in the thumb zone"), and *both* of its mocks put the action in a block at the bottom of the body.

## The slot belongs to the screen, not to one state

Connect's button was one of five state branches, and the `Confirm` branch carried its own "Trust and continue" plus "Cancel" *inside* the certificate card — so "pin the button" was ambiguous in a way it was not on sign-in. It was raised on #155 before building, and the answer is that the slot carries whichever action the current state offers:

| State | Pinned slot |
| --- | --- |
| `Idle` | Connect |
| `Inspecting` | *nothing* |
| `Confirm` | Trust and continue + Cancel, lifted out of the certificate card |
| `Trusted` | Connect (unchanged: transient, the navigation effect leaves it immediately) |
| `Error` | Try again |

Pinning only the idle action would have left the screen worse than untouched: the primary action would then move between the thumb zone, the middle of a card, and the end of the scroller across three states of one screen. The certificate card is now facts only, which is what decision 1 ("the fingerprint is the centerpiece") wants of it anyway.

## It also fixes a live defect — and paid for it with decision 4

The `Confirm` state overflowed at 360×800: the *previous* golden shows "Trust and continue" jammed against the bottom edge and "Cancel" entirely below the fold, on the one screen whose design card says "this is the one decision here".

Pinning alone would have traded that defect for another one. On `main` the card's closing instruction — *"Compare this fingerprint with the one shown by your server before trusting it."*, the copy decision 2 rests on — was on screen at 360×800 and Cancel was not; my first pass swapped which of the two fell below the fold. A review pass caught it, so **decision 4's brand half comes with this change**: the logo, title and subtitle drop away once the certificate shows ("the brand steps aside — that screen belongs to the facts"), which frees the 330 dp that pays for both. At 360×600 the whole card and both actions now fit with nothing to scroll — strictly better than `main` at both heights.

The rest of the mock's confirm layout (its top app bar and "First connection to `<host>`" intro line) stays open; the host is still named by the locked URL field.

## `Inspecting` has no action, and that was tested the hard way

I first built it the way sign-in builds `Submitting` — the Connect button kept in place, disabled — and `AccessibilityChecksTest.connectInspecting` rejected it: contrast 1.88, because M3 renders a disabled label at 38% alpha. That is exactly why `ConnectScreen` already holds the URL field `readOnly` rather than disabled. Sign-in gets away with the same construction only because its submitting button contains a `CircularProgressIndicator` and no text, so there is no label to measure; Connect's wait is already carried by the labelled knot loader. The slot's padding goes with the action, so the state leaves no empty band along the bottom edge rather than a 40 dp hole. Written up on [#155](https://github.com/Xexanos/ratatoskr-app/issues/155#issuecomment-5509332942).

## Tests

- Four Connect goldens re-recorded, plus a new `Connect - confirm, short screen` at 360×600 — the smallest height the design draws for, and the case the 800 dp previews cannot show.
- `AccessibilityChecksTest` green on all 30 checks in **both** light and dark (`cmd uimode night no` / `yes`), including the new preview.
- New `ConnectScreenTest` (9 cases) pins what no golden can see: the state→action mapping, that the brand steps aside in `Confirm`, and — under a `w360dp-h600dp` Robolectric qualifier — that fingerprint, compare hint and both actions are all displayed at once. Its error case asserts the property the slot exists for: on a window short enough to carry the banner off screen, "Try again" is still on it.
- `AppFlowTest` + `MigrationFlowTest` green — the flows tap Connect and Trust through the real navigation.
- `lintDebug`, `verifyRoborazziDebug`, `assembleDebug`, `:app:assembleRelease`, `-PminifiedTests :app:assembleMinifiedAndroidTest`, `scripts/check-ux.sh`, and both release guards all green locally.

## Deliberately not in this PR

Three further gaps on the same design card, now recorded in the card's Δ list instead:

- **The confirm step's top app bar, back button and host intro line** — decision 4's brand half is built (above), its layout half is not.
- The idle mock also pins the URL field and a "Host and port, HTTPS only." helper inside the same bottom block, centring the intro above them. The field stays in the scroll region, as sign-in's fields do, so idle reads as two blocks with a gap where the mock reads as one composition. Worth doing on **both** first-run screens together rather than on Connect alone — fixing it here would re-create the inconsistency this PR removes.
- The fingerprint is still one wrapped string, not four groups of eight.
- `height(52.dp)` on the filled buttons would clip a wrapped label at large font scale; `heightIn(min = 52.dp)` would not. Shared with sign-in, and there is no font-scale preview anywhere in the repo to catch it — also app-wide, also better fixed for both screens at once.

`CONTEXT.md` gains **Pinned action** as a term, since the pattern now spans two screens. It states the wait rule in two branches on purpose: a wait that *is* the action keeps its button and puts the indicator inside it (sign-in submitting, which the design names), while a wait the knot loader owns gives the slot up (Connect inspecting). A single rule would have made the two screens look inconsistent when both are right.

Closes #155


Xexanos:
🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

I ran the `ux-review` skill over this diff before asking for a merge. It found one thing that gated the PR and is now fixed, and left three judgment calls that are yours, not mine. Recording both so the fixed one is auditable and the open ones do not get lost.

**Fixed: pinning traded one below-the-fold defect for another.** On `main`, the certificate card's closing instruction — *"Compare this fingerprint with the one shown by your server before trusting it."* — was on screen at 360×800 while "Cancel" was not. Lifting the actions out of the card put Cancel on screen and pushed that instruction off, on the TOFU screen, where decision 2 ("Explain instead of alarm") rests on exactly that copy. The review's call was that decision 4 stopped being optional the moment the actions were unpinned from the card, and I agree: dropping the logo, title and subtitle in `Confirm` frees 330 dp and now both fit, at 360×600 with nothing to scroll. A `w360dp-h600dp` Robolectric qualifier and the new 600 dp golden pin it.

**Three open questions I deliberately did not decide.**

1. **Is the pinned slot the right home for a blocked-flow retry?** The design says a flow-blocking failure "gets its own full-width button below the banner". I read that as a rule about the retry's *scope* (own button vs. tappable banner) rather than its adjacency, so "Try again" went into the slot — but the banner now ends around y=467 and the button starts at y=732, ~265 dp apart. The alternative is to leave the slot empty in `Error` as it is in `Inspecting` and keep retry attached to its banner. Whichever you pick belongs in `CONTEXT.md`'s new **Pinned action** entry, because it is the rule the next screen copies.

2. **Nothing marks the scroll/slot boundary** — no fade, no divider, no surface behind the slot. It does not bite on Connect any more (nothing scrolls in any Connect golden now), but sign-in's `Sign in - scrolled` golden chops the username field mid-height with no signal that it continues. The design's mocks never have content scrolling under the actions, so there is no rule to cite either way.

3. **Screen margins: 24 dp on the two first-run screens, 16 dp everywhere else,** while the token table says "Screen margins 16 dp". `ThumbZone` now re-declares 24 a fourth time. If 24 is intended for first-run, it belongs in the token table and a shared constant; if not, it is app-wide drift that wants its own issue. Pre-existing either way — I did not touch it.

**Two things I left alone on purpose, both because fixing them here would re-create the inconsistency this PR removes:** idle's composition still reads as two blocks with a gap where the mock centres the intro, and `height(52.dp)` on the filled buttons would clip a wrapped label at large font scale (`heightIn(min = ...)` would not). Both are shared with sign-in and want doing on both screens at once. There is also no font-scale preview anywhere in the repo to catch the second one — worth its own issue if you want that floor.

Xexanos:
🤖 Heads-up from the session that merged #157 (icon weight, #153) — two things will bite this PR on rebase.

**The new UX gate rule will reject the added lock.** `scripts/check-ux.sh` now has a rule 6 that fails on any filled glyph outside the transport set (`PlayArrow|Pause|Stop|Replay10|Forward10|SkipPrevious|SkipNext`). The `Icons.Default.Lock` this PR adds to `ConnectScreen` will be flagged. Switch it to `Icons.Outlined.Lock` (import `androidx.compose.material.icons.outlined.Lock`) — Connect's two existing locks already moved to outlined on main, so this also keeps the screen at one weight.

**One golden conflicts.** #157 re-recorded 26 goldens; this PR touches 5, and `ConnectScreenPreviewsKt.ConnectConfirmPreview.Connect_-_confirm_certificate_W360dp_H800dp.png` is in both sets. Git cannot auto-merge a PNG, so take main's copy and re-record after the rebase:

```
git fetch origin main && git rebase origin/main
git checkout --ours "app/src/test/screenshots/io.github.xexanos.ratatoskr.ui.connect.ConnectScreenPreviewsKt.ConnectConfirmPreview.Connect_-_confirm_certificate_W360dp_H800dp.png"
```

then `:app:recordRoborazziDebug` once the icon switch is in.

Xexanos:
🤖 *Claude Code — agent comment, posted with @Xexanos's token.*

Merged `main` in (#157 plus the 2.2.1 release). Three collisions, all in files both sides had touched — one of them needs your eye because it changes what this PR releases.

**This PR is now 2.2.2, not 2.2.1.** Both branches independently bumped to 2.2.1 and wrote a `20201` changelog. `main`'s is the released one and describes the icon-weight change, so it keeps `20201` verbatim and this branch's entry moved to `20202` with the version behind it. A `fix:` on top of a released 2.2.1 is 2.2.2, and the guard agrees: *"OK: patch-level PR bumps 2.2.1 (20201) -> 2.2.2 (20202); changelog present."* Nothing about the change itself moved — only which release carries it.

**Connect's two Lock glyphs now carry #157's outlined weight** inside the pinned layout this branch builds. `main` had rewritten the same import and both call sites that this branch had restructured around, so git could not merge the hunk. New `check-ux.sh` rule 6 passes on the result.

**The Connect confirm and inspecting goldens were re-recorded by both sides,** so I re-recorded on the merge rather than picking a side — they now carry the outlined lock and the pinned actions together. `verifyRoborazziDebug` is green, and the whole golden set is consistent (main's 20 re-recorded goldens came across untouched).

Re-ran on the merged tree: `verifyRoborazziDebug`, `lintDebug`, `assembleDebug`, `:app:assembleRelease`, `check-ux.sh` (all six rules), both release guards, `AppFlowTest`, and `AccessibilityChecksTest` 30/30 in light **and** dark. All green.

